PR #13, opened 2016-04-02
fix ignoring of keywords inside comments

JeffBezanson:
This was part of the fix for https://github.com/JuliaLang/julia/issues/15461, but I just noticed that it doesn't work, due to a mis-nesting of this clause.

### DIFF
--- a/julia-mode-tests.el
+++ b/julia-mode-tests.el
@@ -284,6 +284,14 @@ end"
     end
 end"))
 
+(ert-deftest julia--test-indent-after-commented-keyword ()
+  "Ignore keywords in comments when indenting."
+  (julia--should-indent
+   "# if foo
+a = 1"
+   "# if foo
+a = 1"))
+
 (defun julia--run-tests ()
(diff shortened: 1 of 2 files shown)


yuyichao:
Add a test?

JeffBezanson:
Test added.

yuyichao:
Hmm, interesting, I tried a few cases (including similar ones) and there's no indentation issues for me. The added test also passes without the patch here.... (this is 24.5.1)...

JeffBezanson:
It should fail in emacs 23.

yuyichao:
I see. Too bad our tests don't run there...

Keno:
Speaking of tests, is there a way we can CI this repository?

yuyichao:
@Keno It is already =) The test doesn't run on emacs 23 though (AFAIK).

Keno:
Never mind, it already is. Just temporarily absent from my GitHub UI for some reason.

PR #18, opened 2016-05-16
Replace {UTF8,ASCII,Byte}String with String

yuyichao:


tkelman:
What's the versioning strategy here? Are there old branches or tags for past versions of Julia?

Issue #9, opened 2016-03-31
Indentation after import/export


yuyichao:
Ref https://github.com/JuliaLang/julia/issues/15197

> julia-mode.el does not indent after import/export according to how the source is indented. For example, it will *not* indent the following code this way.
>
> ```julia
> import ..LinAlg: BlasFloat, Char, BlasInt, LAPACKException,
>     DimensionMismatch, SingularException, PosDefException, chkstride1, checksquare
> ```
>
> Specifically, there is no indentation on the second line

yuyichao:
Should be closed by #20

PR #20, opened 2016-07-31
Add indent for import and export.

tpapp:
Should fix issue #9.

Tried to follow the style of existing code, but I am not familiar with Emacs Lisp, please look at implementation.


tkelman:
same would be appropriate for `using` if that were to get long mentioning specific methods

yuyichao:
LGTM. Thanks.

PR #21, opened 2016-08-01
`python-mode`-like paren-indent

kshramt:
This commit provides the following indentation-style, which is similar to that of `python-mode`:

```julia
function1(a, b, c
          d, e, f)
function1(a, b, c
          d, e, f
          )
function2(
    a, b, c
    d, e, f)
function2(
    a, b, c
    d, e, f
)


for i in Float64[1, 2, 3, 4
                 5, 6, 7, 8]
end
for i in Float64[1, 2, 3, 4
                 5, 6, 7, 8
                 ]
end
for i in Float64[
    1, 2, 3, 4
    5, 6, 7, 8]
end
for i in Float64[
    1, 2, 3, 4
    5, 6, 7, 8
]
end


a = function3(function ()
              return 1
              end)
a = function3(function ()
              return 1
              end
              )
a = function4(
    function ()
    return 1
    end)
a = function4(
    function ()
    return 1
    end
)
```

### DIFF
--- a/julia-mode-tests.el
+++ b/julia-mode-tests.el
@@ -183,6 +183,29 @@ baz )"
 foobar( bar,
         baz )"))
 
+(ert-deftest julia--test-indent-paren-newline ()
+  "python-mode-like indentation."
+  (julia--should-indent
+     "
+foobar(
+bar,
+baz)"
+     "
+foobar(
+    bar,
+    baz)")
+  (julia--should-indent
+     "
+foobar(
+bar,
+baz
+)"
+     "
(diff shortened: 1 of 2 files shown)


yuyichao:
Travis don't want to test this?

yuyichao:
The new bahavior LGTM. The indentation of `function` (or blocks) in parenthesis is still bad but that's not a regression and has a separate issue.

kshramt:
Thank you for your comment.
Yes, indentation inside the block is not a regression.

Issue #11, opened 2016-03-31
Indentation inside parenthesis


yuyichao:
Ref https://github.com/JuliaLang/julia/issues/8971
Ref https://github.com/JuliaLang/julia/issues/11498

A few cases right now.

```jl
function1(a, b, c
          d, e, f)
function2(
          a, b, c
          d, e, f)
for i in Float64[1, 2, 3, 4
                 5, 6, 7, 8]
end
for i in Float64[
                 1, 2, 3, 4
                 5, 6, 7, 8]
end
a = function3(function ()
              return 1
              end)
a = function4(
              function ()
              return 1
              end)
```

Some proposed changes

```jl
function1(a, b, c
          d, e, f)
# Or
function1(a, b, c
    d, e, f)
function2(
    a, b, c
    d, e, f)
for i in Float64[1, 2, 3, 4
                 5, 6, 7, 8]
end
# Or
for i in Float64[1, 2, 3, 4
        5, 6, 7, 8]
end
for i in Float64[
        1, 2, 3, 4
        5, 6, 7, 8]
end
a = function3(function ()
                    return 1
                end)
a = function4(
                function ()
                    return 1
                end)
```

christopher-dG:
I think this is relevant:
```julia
longfunctionname(;  # Wrong, in my opinion
                 hello=nothing,
                 world=nothing,
                 )
longfunctionname(  # Right
    hello,
    world,
)
```

non-Jedi:
Current state:

``` julia
function1(a, b, c,
          d, e, f)
function2(
    a, b, c,
    d, e, f)
for i in Float64[1, 2, 3, 4
                 5, 6, 6, 8]
end
for i in Float64[
    1, 2, 3, 4,
    5, 6, 7, 8]
end
a = function3(function ()
              return 1
              end)
a = function4(
    function ()
    return 1
    end)
```

With the possible exception of `function3` and `function4`, this
matches my expectation and how I would indent code if I were doing so
manually.

ronisbr:
I think the major problem here is that the indentation is completely lost in cases like:

```julia
        hl = Highlighter((h,data,i,j)->begin
                         id = div(i-1, 3)
                         sim_id = id*num_cols + j
                         if sims_status[sim_id] == 1
                         return crayon"black bg:green"
                         else
                         return crayon"bg:yellow"
                         end
```

FelipeLema:
[...]

PR #25, opened 2016-09-10
Indent block inside parenthesis

kshramt:


yuyichao:
Fixes https://github.com/JuliaLang/julia-emacs/issues/11 ?

Maybe also a test for the case where there's no new line between `(` and a block start? (`function3` case in #11)

kshramt:
This patch fixes only `function4` of #11.
Fixing `function3` of #11 may require much more work.
As I do not write code in `function3` style, I do not have strong motivation to write patch for `function3`...

kshramt:
This patch does not indent the following case correctly:

```julia
function f(
    x
    )
end
```

Issue #4, opened 2016-03-31
Improve highlighting of quoted symbols


yuyichao:
Ref https://github.com/JuliaLang/julia/issues/9247

> > I think it would make sense for quoted symbols to use constant-face by default; they are constants after all.
> > 
> > Also quoted-symbol-face seems not to work right at the beginning of a line. This is definitely a pretty rare case though.

PR #28, opened 2016-10-07
font-locking for symbol at beginning of line

nverno:
hi, added `bol` to regex for symbol so 
`:a in ...` gets font-locked

### DIFF
--- a/julia-mode-tests.el
+++ b/julia-mode-tests.el
@@ -54,6 +54,16 @@
        (should (equal (buffer-substring-no-properties (point-min) (point-max))
                       ,to)))))
 
+(defmacro julia--should-font-lock (text pos face)
+  "Assert that TEXT at position POS gets font-locked with FACE in `julia-mode'."
+  `(with-temp-buffer
+     (julia-mode)
+     (insert ,text)
+     (if (fboundp 'font-lock-ensure)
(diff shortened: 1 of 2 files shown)


yuyichao:
Is there a way to test this?

nverno:
If you evaluate it with `M-C-x` and reload file.  Or do you mean a more rigorous test?

yuyichao:
I mean something similar to https://github.com/JuliaEditorSupport/julia-emacs/blob/master/julia-mode-tests.el that can be run automatically and on the CI.

yuyichao:
(Given that we don't have a similar test now I'm fine with merging even if no one knows how to add a similar test in a few days but it's still better to have a test for that.)

nverno:
Ok thanks, I was just thinking about what kind of test to make for font-locking, but I don't know enough of julia's syntax to know what I would be testing for really.

yuyichao:
You don't need to add tests for all the supporting syntax (you are welcome to of course). For this PR, if you know how such things can be tested (make sure user settings are overwritten for the test) simply add a test for the case where it didn't highlight correctly before but does with this PR should be good enough.

nverno:
I think that test covers this case.  Related to #4

yuyichao:
Why is this closed?

PR #32, opened 2016-10-30
fully functional julia repl in emacs

gaika:
create fully functional repl in emacs using screen and term mode. send line or buffer from to repl for evaluation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,3 +10,11 @@ Put the following code in your `.emacs`, `site-load.el`, or other relevant file
 (add-to-list 'load-path "path-to-julia-mode")
 (require 'julia-mode)
 ```
+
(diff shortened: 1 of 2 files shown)


gaika:
just found julia-repl https://github.com/tpapp/julia-repl

tpapp:
If you think that your solution implements something missing from https://github.com/tpapp/julia-repl (which is still experimental), please let me know, I would be happy to cooperate.  PRs are welcome.

gaika:
The only difference is that julia process is run under screen which gives better terminal emulation. There's no need to modify global term properties and all keys work out of the box.

tpapp:
@gaika: Thanks for the suggestion. I added an option for `julia-repl` to use screen when available, in the latest master.

Issue #27, opened 2016-10-07
wrapping control statements in parenthesis leads to wrong indentation


jshysing:
I prefer to wrap control statements in parenthesis, however this seems to break the indentation like in the following example ( applying end() seems to fix the issue but that's obviously not a nice fix)

```julia
if( a>0 )
end

    for( i=1:2 )
        for( j=1:2 )
            for( k=1:2 )
            end
            end
        end

        end

```

yuyichao:
Likely dup of https://github.com/JuliaEditorSupport/julia-emacs/issues/10

PR #35, opened 2017-02-09
Fix indentation for anonymous functions.

### LINKED ISSUE #10: Indent after anonymous function
Ref https://github.com/JuliaLang/julia/issues/15691

> It seems that the indent after an anonymous function without a space between `function` and the opening parentheses is wrong.
> 
> ``` jl
> function()
> end
>     b = 1
> ```

tpapp:
This should fix issue #10. The problem was that when the indentation
algorithm traversed forms with `function(args...)`, and it was on the
`(`, `(julia-at-keyword julia-block-start-keywords)` would return `t`
because `(current-word t)` would return `function`.

The fix adds a check for this. Tests are also added.

Fix #10

### DIFF
--- a/julia-mode-tests.el
+++ b/julia-mode-tests.el
@@ -374,6 +374,27 @@ end"
     end
 end"))
 
+(ert-deftest julia--test-indent-keyword-paren ()
+  "indentation for ( following keywords"
+  "if( a>0 )
+end
+    
+    function( i=1:2 )
+        for( j=1:2 )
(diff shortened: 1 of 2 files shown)


yuyichao:
Is this specific to functions? Or does it fix https://github.com/JuliaEditorSupport/julia-emacs/issues/27 as well?

tpapp:
Good catch! It does appear to fix #27, and I added tests for that.

PR #36, opened 2017-02-10
update for changed type keywords in 0.6

JeffBezanson:


julia-mode.el
```diff
@@ -240,7 +240,9 @@ This function provides equivalent functionality, but makes no efforts to optimis
       (not (any "="))))
 
 (defconst julia-type-regex
-  (rx symbol-start (or "immutable" "type" "abstract") (1+ space) (group (1+ (or word (syntax symbol))))))
+  (rx symbol-start (or "immutable" "type" ;; remove after 0.6
+                       "abstract type" "primitive type" "struct" "mutable struct")
```
Wilfred: This assumes that users will never write `abstract␣␣type`. `(or "immutable" "mutable" "type" "abstract" "primitive" "struct")` might be a little more robust.

JeffBezanson: In the future, `type` won't be a keyword anymore. How about `(seq "abstract" (1+ space) "type")`?

Wilfred: That would certainly work :+1:

PR #37, opened 2017-03-12
Fix backslash character class.

### LINKED ISSUE #17: parenthesis matching and indentation of `\)`
``` jl
(\)
 1
 (:\)
  1
```

The `(\)` and `(:\)` are both complete expressions and it shouldn't indent the line afterward. In general I think `\` doesn't escape anything outside string in julia.

tpapp:
Fixes #17, with tests.


yuyichao:
(Please feel free to ping. I generally wait a day or so before merging but will sometimes forget about  it.......................)

PR #39, opened 2017-04-25
add support for `where` keyword

rfourquet:


yuyichao:
Can the two changes be tested?

rfourquet:
Done, except that I don't test for the necessary absence of space addressed in the the second commit (don't know how to do!).

PR #42, opened 2017-05-14
Insertion of math symbols in julia using auctex's mechanism

antoine-levitt:
This enables the insertion of mathematical symbols in Julia using the latex-math-mode mechanism of a prefix key. I'm not sure what the preferred mechanism is for an emacs package that depends on another one, here I just (require latex) and check for errors.


antoine-levitt:
Any appetite for merging this?

antoine-levitt:
Thanks!

Issue #43, opened 2017-07-07
adding binary operator special characters


PinkMushroom:
Hi,
could we make \xor a synonym for \veebar ?
also, maybe consider the following
\bigcap \and => map to 'and' function
\bigcup \or => map to 'or' function

I actually tried to do a pull requesst for this but couldn't due to denied permission when i tried to push my changes.
do you need an account to work on julia emacs mode ?

jballanc:
> do you need an account to work on julia emacs mode ?

You will need a GitHub account, which is the same as the account you used to file this issue.

The usual way of submitting pull requests is to first "Fork" the repository (it's the button in the upper right corner). This will give you your own private "copy" of the `julia-emacs` repository. Using this repository you can then create a branch, push your changes, and open a pull request to have that branch merged back into the main repository.

PR #44, opened 2017-07-10
Bin op chars

PinkMushroom:
very simple changes to assign special chars to binary operators
xor should map to veebar
and should map to bigand
or should map to bigor

Fix https://github.com/JuliaEditorSupport/julia-emacs/issues/43


PinkMushroom:
made  requested changes

PR #45, opened 2017-09-15
Regenerate latex auto completion

yuyichao:


julia-mode.el
```diff
@@ -907,6 +907,7 @@ strings."
 (puthash "\\gamma" "γ" julia-latexsubs)
 (puthash "\\delta" "δ" julia-latexsubs)
 (puthash "\\upepsilon" "ε" julia-latexsubs)
+(puthash "\\varepsilon" "ε" julia-latexsubs)
```
yuyichao: I'm not sure if it's just me but this renders differently from what I see locally....

spectacle x20264
spectacle l20315

yuyichao: The previous one renders the way I expect it to in the browser though...

spectacle x20422

PR #47, opened 2017-11-15
Fixed import export etc indentation at start of buffer.

tpapp:
Previously julia-mode would indent these keywords at the start of the
buffer, because the code had an error (point can never go below
point-min, which is at least 1). Fixed this, fixed tests (which
started with a blank line so did not catch this).


yuyichao:
I assume this is why the first line of import will toggle between indented and not indented?

tpapp:
Yes, it fixes that one.

Issue #48, opened 2017-12-20
Many Unicode characters are missing from tab completion


saweiss:
I noticed that many Unicode characters are not available for tab completion. The latest list of available characters can be found [here](https://docs.julialang.org/en/latest/manual/unicode-input/). Can anyone help update the completion list? I will try to add some when I can. Thanks!

vspinu:
BTW, is it really necessary to keep julia's own unicode list and completion as part of the repo? There are a few quite comprehensive lists in [`math-symbol-lists`](https://github.com/vspinu/math-symbol-lists) and there is [`company-math-symbols-unicode`](https://github.com/vspinu/company-math) backend which could be used globally not just for julia mode. 

If julia doesn't understand some symbols a few false-positives are not a big deal, right?

non-Jedi:
Should I make a PR for the local changes I have that expands the
completions to also include valid Julia emoji (e.g. `\:watermelon:` =>
🍉), or is that a Very Bad Idea for reasons?

It's literally a single change that merges `REPL.REPLCompletions.emoji_symbols` with `latex_symbols`.

yuyichao:
That sounds find to me. I'll also close this since I assume the generic problem is fixed by #67 already.

PR #49, opened 2017-12-20
Add more Unicode characters for tab completion

saweiss:
Add Unicode characters from \:mahjong: to \:stars: for tab completion. See #48 for more details.


tpapp:
I think this is superseded by #67, which makes generation easier, and following PRs like #78, and could be closed.

Issue #15, opened 2016-04-15
Highlight of triple quote string with escaped quote


yuyichao:
The triple quote string is not highlighted correctly if there's another triple quote in the string that has only the first quote escaped.

Screenshot,
![spectacle x15522](https://cloud.githubusercontent.com/assets/712232/14566350/422393c4-02fd-11e6-8b46-33592f82b943.png)
The first line is highlighted correctly but the second one not.

rfourquet:
Related: highlight of a single quote string with ecaped quotes fails: `"a\"b"`.

christopher-dG:
Also related, all text following this string is highlighted as part of the string:

`"""he said: "hello, world\""""`

PR #54, opened 2018-04-29
Handle string escapes

jamii:
Before:

Before

After:

After

### DIFF
--- a/julia-mode-tests.el
+++ b/julia-mode-tests.el
@@ -435,6 +435,13 @@ end")
       end"
      pos 'font-lock-keyword-face)))
 
+(ert-deftest julia--test-escaped-strings-dont-terminate-string ()
+  "Symbols get font-locked at beginning or line."
+  (let ((string "\"\\\"\"; function"))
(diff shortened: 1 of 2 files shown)


jballanc:
Honest question: is there a way to write tests for highlighting like this in Emacs?

jamii:
I'm sure I saw some tests when I was looking for examples for this patch, but I can't find them again.

yuyichao:
I believe the tests are here https://github.com/JuliaEditorSupport/julia-emacs/blob/344e6c2a32b528eef87165ac32151990193a70f1/julia-mode-tests.el#L406

jamii:
Now there is a test :)

jamii:
Note this doesn't fix https://github.com/JuliaEditorSupport/julia-emacs/issues/15

yuyichao:
> Note this doesn't fix #15

Yeah, I know. I assume it does fix #55 though, which was a dup of the issue mentioned in https://github.com/JuliaEditorSupport/julia-emacs/issues/15#issuecomment-315607776.

PR #66, opened 2018-12-12
Fix for derived parent mode

yuhan0:
The parent major mode of julia-mode is registered as the alias `julia-mode-prog-mode`,  leading to checks such as  

```emacs-lisp
(derived-mode-p 'prog-mode)
```

returning nil. This PR adds a hack after the mode is defined to overwrite the `derived-parent-mode` prop of the mode symbol.  

A better solution might be to drop support for Emacs 23 and derive directly from `prog-mode`? 

For context Emacs 24 was released in 2012, around the initial Julia 0.2 release, and I doubt there are many users of older Emacs versions out there to justify the compatibility workarounds throughout the code. ..

### DIFF
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -635,11 +635,6 @@ meaning always increase indent on TAB and decrease on S-TAB."
     (when (>= point-offset 0)
       (move-to-column (+ (current-indentation) point-offset)))))
 
-(defalias 'julia-mode-prog-mode
-  (if (fboundp 'prog-mode)
-      'prog-mode
-    'fundamental-mode))
-
 
 ;;; Navigation
 ;; based off python.el
@@ -781,7 +776,7 @@ Return nil if point is not in a function, otherwise point."
     ))
 
 ;;;###autoload
-(define-derived-mode julia-mode julia-mode-prog-mode "Julia"
+(define-derived-mode julia-mode prog-mode "Julia"
   "Major mode for editing julia code."
   (set-syntax-table julia-mode-syntax-table)
   (set (make-local-variable 'comment-start) "# ")
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -367,14 +367,13 @@ Based on `python-syntax-stringify'."
                            'syntax-table (string-to-syntax "|"))))))
 
 (defconst julia-syntax-propertize-function
-  (unless (< emacs-major-version 24)
-    (syntax-propertize-rules
-     ("\"\"\""
-      (0 (ignore (julia-stringify-triple-quote))))
-     (julia-char-regex
-      (1 "\"")                 ; Treat ' as a string delimiter.
-      (2 ".")                  ; Don't highlight anything between.
-      (3 "\""))))) ; Treat the last " in """ as a string delimiter.
+  (syntax-propertize-rules
+   ("\"\"\""
+    (0 (ignore (julia-stringify-triple-quote))))
+   (julia-char-regex
+    (1 "\"")                    ; Treat ' as a string delimiter.
+    (2 ".")                     ; Don't highlight anything between.
+    (3 "\"")))) ; Treat the last " in """ as a string delimiter.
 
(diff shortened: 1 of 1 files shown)


tpapp:
I think that dropping support for Emacs 23 would be the way to go. Emacs 24 was released more than 6 years ago, even someone using some ancient LTS distro should have it.

nverno:
We could also just bump the version, so users of old emacs could use older versions. There is quite a bit of compatibility annoyance having to support 23.

yuhan0:
Sounds good to me :) I'll amend this PR to use prog-mode directly if that's alright

yuyichao:
> For context Emacs 24 was released in 2012, around the initial Julia 0.2 release, and I doubt there are many users of older Emacs versions out there to justify the compatibility workarounds throughout the code. ..

> I think that dropping support for Emacs 23 would be the way to go. Emacs 24 was released more than 6 years ago, even someone using some ancient LTS distro should have it.

None of these actually matters. Old emacs users should not be bound to use old julia versions and different software have very different support timeline. Ubuntu is also not the only relavant distro....

The oldest emacs version that should be supported should be at least the default emacs on the oldest environment the latest julia supports. We still support CentOS 6 (and AFAICT for whatever reason people might still be using it...) which seems to ship emacs 23.1 so emacs 23 should be a supported version and it still makes sence to include update for new julia features for that emacs version.

Of course it doesn't have to be supported with the same code, branching for emacs 23 is no problem and integration with emacs 24 feature certainly doesn't need to be backported to emacs 23. Update related to new julia features though, should.

tpapp:
In principle I would be fine with keeping support for Emacs 23, but Emacs 24 introduced lexical scope which makes things a lot easier *and* faster. AFAIK [there is no way to use it while supporting both 23 and 24](https://stackoverflow.com/questions/12456479/lexical-scope-in-emacs-compatibility-with-older-emacsen), so many packages dropped support for Emacs 23. Cf #61.

Also, while I am familiar with CentOS [emacs 24.3 seems to be available](https://centos.pkgs.org/6/puias-unsupported-x86_64/emacs24-24.3-11.sdl6.1.x86_64.rpm.html) in the PUIAS repo. There are always trade-offs, but I imagine that a *bit* of extra effort from users who want to use something from 2011 in exchange for making the task of maintainers easier is not a bad one.

yuyichao:
Having a newer version from a non-default (I assume) repo doesn't change thing by too much since I think people that want to use julia on centos6 probably didn't pick the OS by their own choice and doesn't have root access to install systemwise package.....

Anyway, as I said, I have no problem splitting the support for emacs 23 and 24 into different branches. It's just that the release time of the new version isn't the deciding factor here.

yuhan0:
Is there anything I can do to improve on this PR? 

The corresponding fix for Emacs 23 was to add the following, if a separate branch is to be created:
```elisp
;; necessary for `derived-mode-p' to work with the julia-mode-prog-mode alias.
(put 'julia-mode 'derived-mode-parent
     (if (fboundp 'prog-mode)
         'prog-mode
       'fundamental-mode))
```

Sorry for sidetracking the discussion by mentioning release dates - it was just an observation that the intersection between Julia and Emacs <=23 users would be very small and cautious of upgrading packages, given how many major packages now only support Emacs 25 at minimum.

yuhan0:
Can this be considered again now that Emacs 23 support is officially dropped? 

(there's a trivial conflict with the 2nd commit which wasn't directly related to the main purpose of this PR - let me know if I should drop the commit or rebase it)

tpapp:
Yes, of course, this would be a nice improvement for the code (sorry that it took so long to drop Emacs 23).

I would greatly appreciate if you could refresh the PR, eg by merging in `master` and resolving the conflicts, or making a new one (whichever is easier).

yuhan0:
I've rebased the branch onto master and resolved conflicts.

tpapp:
@giordano, @non-Jedi, if either of you have the time a review would be appreciated.

PR #67, opened 2018-12-27
Move LaTeX substitutions to a separate file, clean up generating script.

tpapp:


tpapp:
This overlaps with/complements #64, if that is merged first I will of course update this.

PR #69, opened 2019-02-12
Update keywords and builtin types to Julia 1.0

emmt:
- Add \prime to the list of word constituent.
- Fix/replace a number of keywords and builtin types for Julia 1.0 syntax.
- Add `Cstring` and `Cwstring` to the list of builtin types.
- Add `undef` to the list of font lock keywords.


tpapp:
Also I think `Associative` is now `AbstractDict`.

christopher-dG:
Bump!

PR #70, opened 2019-03-10
Make julia-latexsub work if point in the middle of latex symbol.

adam-higuera:
This pull request fixes two minor issues I had:

1. emacs was erroring off because the `invoke-command` machinery was calling the zero-argument function `julia-latexsub` with a nil argument.  Removing the `i` in the argument specifier string fixed this (might be a Spacemacs issue, not a julia-mode issue).
2. It wasn't doing the substitution if point is in the middle of the latex command string.

The second change seems relatively innocuous, the first might break things that invoke `julia-latexsub` in ways that Spacemacs doesn't, i.e. with an argument (although it hasn't broken anything for me)

### DIFF
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -819,18 +819,20 @@ strings."
 
 (defun julia-latexsub ()
   "Perform a LaTeX-like Unicode symbol substitution."
-  (interactive "*i")
-  (let ((orig-pt (point)))
+  (interactive "*")
+  (let ((orig-pt (point))
+        (word-end (progn (forward-word)
+                         (point))))
     (while (not (or (bobp) (= ?\\ (char-before))
 		    (= ?\s (char-syntax (char-before)))))
       (backward-char))
     (if (and (not (bobp)) (= ?\\ (char-before)))
         (progn
           (backward-char)
-          (let ((sub (gethash (buffer-substring (point) orig-pt) julia-latexsubs)))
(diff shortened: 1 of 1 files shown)


rfourquet:
Unless I made a mistake, this doesn't work well for me: the substitution happens only if my point is within the word, not when it's at the end (in other words, the only thing which doesn't work is what was working before!). Can you confirm?

julia-mode.el
```diff
@@ -819,18 +819,20 @@ strings."
 
 (defun julia-latexsub ()
   "Perform a LaTeX-like Unicode symbol substitution."
-  (interactive "*i")
-  (let ((orig-pt (point)))
+  (interactive "*")
+  (let ((orig-pt (point))
+        (word-end (progn (forward-word)
+                         (point))))
```
tpapp: The problem is here, it goes forward one character too much.

tpapp: I think this will make it work
```suggestion
                         (backward-char)
                         (point))))
```

tpapp:
I think I can make it work with the suggestion above (currently testing it in practice).

But in general, I am a bit bothered by the fact that we don't have any unit tests for completion, so I will contribute some and in the meantime hold off on this PR.

tpapp:
When #81 is merged, it should also be merged to this branch, and then some tests can be added (or commented out).

non-Jedi:
Superseded by #185

PR #74, opened 2019-05-21
Update substitutions to include emoji

non-Jedi:
As promised in #48, I merged `emoji_symbols` and re-ran `make-julia-latexsubs.jl`. Of course, now that filename is technically lying since the emoji subs aren't really latex subs as far as I know, but that should be fine.

I ran the script with Julia 1.1.1. If any symbols have been newly been made available with 1.2, we'll have to re-run it then.


non-Jedi:
@yuyichao ping

non-Jedi:
thanks

PR #75, opened 2019-05-29
Don’t stop at empty lines in ‘end-of-defun’

dvdhansen:


julia-mode.el
```diff
@@ -759,7 +759,8 @@ Return nil if point is not in a function, otherwise point."
         (while (and (not (eobp))
                     (forward-line 1)
                     (or (julia-syntax-comment-or-string-p)
-                        (> (current-indentation) beg-defun-indent)))))
+                        (> (current-indentation) beg-defun-indent)
+                        (looking-at-p "^\\s-*\\(?:#.*\\)*$")))))
```
nverno: should this just use `forward-comment`?

dvdhansen: We also need to move over empty lines including those with “only but
not enough” whitespaces.  `forward-comment` won’t do the job.
Replacing `looking-at-p` with `(forward-comment 1)` will move the
cursor to the line with the `return` statement and return `nil`.

```julia
function f()

    x = 42
    return x
end
```

nverno: There is a bug in cases like

```
f(x)::Int8 =
# no comment
    x*x

function forty_two()
 # no comment
    return 42
end
```

where `end-of-defun` jumps from first to end of second.  It also gets stuck when point is just prior to the first function.

nverno: For `forward-comment` I had been thinking along the lines of

```
(while (and (not (eobp))
                    (forward-line 1)
                    (or (forward-comment 5)
                        (julia-syntax-comment-or-string-p)
                        (> (current-indentation) beg-defun-indent)))))
```

PR #81, opened 2019-11-08
Add unit tests for LaTeX substitution.

tpapp:


julia-mode-tests.el
```diff
@@ -595,6 +595,32 @@ end
...
+    (insert contents)
+    (goto-char position)
+    (julia-latexsub)
+    (buffer-string)))
+
+(ert-deftest julia--test-substitutions ()
+  (should (equal (julia--substitute "\\alpha " 7) "α "))
+  (should (equal (julia--substitute "x\\alpha " 8) "xα "))
+  (should (equal (julia--substitute "\\kappa\\alpha(" 13) "\\kappaα("))
+  ; (should (eq (julia--substitute "\\alpha" 7) "α")) ; BROKEN
```
rfourquet: Did you use `eq` on purpose here? It indeed fails as is, but passes with `equal`. On the other hand, the next test (with position 6) fails even with `equal`,

tpapp: No, it just a mistake. I updated the code. Thanks!

PR #84, opened 2019-11-11
Run CI using emacs 25

non-Jedi:


tpapp:
AFAIK Travis has bionic:

https://docs.travis-ci.com/user/reference/bionic/

Can you please try that instead of using a PPA?

non-Jedi:
We'd lose testing on 23 then I think. To my mind, if we're supporting old emacs that's much more likely to break without anyone noticing than 25/26. One alternative is that we could use flycheck's setup instead? https://github.com/flycheck/emacs-travis

tpapp:
I think that dropping support for Emacs 23 would be the way to go. I opened a discussion in #86.

tpapp:
Thanks for calling attention to this. Closed this in favor of #95, but if you want to take a stab at setting up CI with emacs-travis, a PR would be appreciated.

PR #85, opened 2019-11-11
WIP: Fix backslash-indentation and dotted-operator indentation

### LINKED ISSUE #56: wrong indentation with `.|>`
Should be
```julia
x .|>
   f
```
is
```julia
x .|>
f
```

### LINKED ISSUE #58: \( upsets paren matching
screenshot_2018-12-03_09-27-06

Using `julia-emacs` `master`, Emacs 26.1.

non-Jedi:
Closes #58. Closes #56.

There may be a better way of handling dotted operators, but I went with matching what was already there.


non-Jedi:
I missed the bit where `\` has to be an escape char for string escaping to work correctly. For reference of how to do this correctly see `octave-syntax-propertize-sqs`. Don't have time to fix this evening.

non-Jedi:
The fix here is dependent on dropping Emacs 23 support (#86). Not going to implement twice, once with `syntax-propertize-function` and once without.

tpapp:
This is fine, I will just wait a month for feedback before merging #86.

non-Jedi:
This is going to end up looking entirely different when I get around to it. Will open a new PR.

tpapp:
Great, thanks.

PR #87, opened 2019-11-13
Transition to cl-lib, drop support for Emacs 23.

### LINKED ISSUE #86: RFC: drop support for Emacs 23
Supporting only Emacs 24 and above would have the following advantages:

1. simpler CI setup (cf #84)

2. ability to use `cl-lib` (instead of `cl` as we currently do, cf #60).

Please comment on this, especially if you need Emacs 23. As developers are usually on the bleeding edge, it is easy to forget about users operating in legacy environments.

Some data:

- Emacs 24.1 was released on 2012-06-10,

- the oldest Ubuntu LTS supporting Emacs 24 is Trusty (2014-04), other distributions have a similar timeline

tpapp:
Incidental changes:

- use lexical binding (requires Emacs 24)

- rename a variable `map` which was shadowing the eponymous function (innocuous here, but bad style)

- bump version

Fixes #86. Relevant also for #84.


rfourquet:
Sorry to react so late, thanks for this change!

PR #92, opened 2019-12-02
try to fix loading

### LINKED ISSUE #91: Looks for julia-latexsubs in directory of file in current buffer
On julia-mode 20191128.1452, if (for example) I open the file
`/home/username/.julia/dev/MyPackage/src/foo.jl`
I get the following error:

`File mode specification error: (file-missing Cannot open load file No such file or directory /home/myusername/.julia/dev/MyPackage/src/julia-latexsubs)`

After this, there is no syntax highlighting or other expected functionality.

This is on Cygwin. I could try Linux later if no one can reproduce on Linux.

tpapp:
cf https://github.com/JuliaEditorSupport/julia-emacs/pull/83#issuecomment-560211950,
fixes #91 (hopefully), thanks @AlexLewandowski

This was broken by #83. @nverno, can you check that it still solves your original issue?

@chriselrod, can you check that it fixes #91 for you?

### DIFF
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -854,7 +854,10 @@ strings."
 ;;; populate LaTeX symbols hash table from a generated file.
 ;;; (See Julia issue #8947 for why we don't use the Emacs tex input mode.)
 (load (expand-file-name "julia-latexsubs"
-                        (file-name-directory (or buffer-file-name load-file-name))))
+                        (file-name-directory
+                         (if load-in-progress
+                             load-file-name
+                           buffer-file-name))))
 
 ;; Math insertion in julia. Use it with
 ;; (add-hook 'julia-mode-hook 'julia-math-mode)


chriselrod:
I'm not sure if I tested this correctly.
I cloned this repo into `/home/username/.emacs.d/sourceinstalls`, and then had the following in my `init.el`:
```elisp
(add-to-list 'load-path "/home/username/.emacs.d/sourceinstalls/julia-emacs")
;; (require 'julia-mode)
(require 'julia-repl)
(add-hook 'julia-mode-hook 'julia-repl-mode)
```
The package did not seem to load (no errors, no syntax highlighting, no latex completions). If installed from melpa, instead of using the `add-to-list`, this resulted in the julia-latexsubs error. It works if I add `(require 'julia-mode)`, but that also fixes the julia-latexsubs error.

non-Jedi:
Why are we messing about with `load` anyway? Why not just make `julia-latexsubs.el` `(provide 'julia-latexsubs)` and define the constant `julia-latexsubs` and then just require it from `julia-mode.el`? As far as I can tell, that would fix these issues and would be simpler than current setup.

nverno:
Yea, this works for `eval-buffer` -- I would still vote for the initial `(when load-file-name ...)` change so you don't have to account for all the possible scenarios, or turn it into a provided feature -- eg. does this always work during byte-compilation (it seems to from a quick check, but I don't know when/if `load-file-name`/`buffer-file-name` are defined in that case)?

tpapp:
I think that the solution using features (#93) is much cleaner than this one.

tpapp:
Closing in favor of #93.

PR #93, opened 2019-12-03
Load latexsub table as a feature

### LINKED ISSUE #91: Looks for julia-latexsubs in directory of file in current buffer

non-Jedi:
Fixes #91. Alternative to #92.

Variable renamed from `julia-latexsubs` to `julia-mode-latexsubs` as incremental improvement towards melpa guidelines.


tpapp:
I think this may be the cleanest solution, favoring this over #92.

tpapp:
My preliminary exploration suggests that this should work fine with

1. MELPA, simply including all the `.el` files like now, without an extra recipe,

2. the Emacs package loading mechanism, which puts package directories in `load-path` so it should always fine the relevant file.

I will keep testing this, comments and reviews are welcome.

nverno:
@non-Jedi If you are still considering going the abbrev route for latex substitutions, I forgot to mention I have been doing similar in my personal configs, although I believe the generating script to produce the abbrev table from the unicode generating script is out-of-date.

If your interested, the relevant script to generate the abbrev table is at [latex_abbrevs.jl](https://github.com/nverno/nvp/blob/master/bin/latex_abbrevs.jl) which should produce a [unicode-latex-abbrev-table](https://github.com/nverno/nvp/blob/master/abbrevs/unicode-latex-abbrev-table).

I'm also a bit obsessive about abbrevs, so there are lot of various abbrev hacks in there.  A really nice feature that comes with have a unicode latex substitution table, is the ability to use it in any other mode.  The method I found the most useful and least intrusive (only minimally effecting the other active abbrev tables by giving them precedence, and being easily disable/enabled on the fly) was have an interactive function to add an additional abbrev table to the current buffer by adding it as the parent of the `local-abbrev-table`.  Adding the unicode expansions on the fly is nice if you know when they'll be useful or annoying.  This is the [function](https://github.com/nverno/nvp/blob/ad61ab4013ed838bb36f2684fef538a29d77b413/nvp-abbrev.el#L177) I'm referring, and is quite simple.  Sadly, I have a couple personal functions worked in there that could be annoying to deal with.

PR #96, opened 2019-12-21
add a changelog

tpapp:


tpapp:
I think that a CHANGELOG is pretty uncontroversial as good practices go, so I am merging this.

PR #100, opened 2020-02-29
Use completion-at-point and abbrev-mode for LaTeX substitution

non-Jedi:
## Disclaimer

Additional commits have been pushed to this branch since this original post was written. This PR now implements immediate substitution of the Unicode symbol for the TeX string upon completion instead of requiring entering a space or symbol or manually triggering `expand-abbrev` which is shown and described here. [This comment is my most up to date summary of the pros and cons of this change](https://github.com/JuliaEditorSupport/julia-emacs/pull/100#issuecomment-1339770633).

## Original Post

I think this is a more sane version of #82. When `|` represents the cursor, all of `\l|am`, `\lam|`, and `\lam|someothertext` will complete as expected. The latex backslash-string won't be automatically substituted for the unicode symbol unless the user inserts a whitespace or calls `expand-abbrev` <kbd><kbd>C</kbd>-<kbd>x</kbd> <kbd>a</kbd> <kbd>e</kbd></kbd>. Ping @nverno as being similar to your personal setup.

Summarizing this PR:

1. Under `julia-mode` `completion-at-point` will now complete LaTeX symbols to their full text for both the case of cursor in the middle of the symbol and cursor at the end (`\la|m` -> `\lambda` and `\lam|othertext` to `\lambdaothertext`).
2. <kbd>TAB</kbd> in `julia-mode` now first tries to indent the line, and if already indented uses `completion-at-point` to complete LaTeX strings.
3. `expand-abbrev` (<kbd><kbd>C-x</kbd> <kbd>a</kbd> <kbd>e</kbd></kbd>) or entering a non-word-syntax character after a LaTeX string will now replace the LaTeX string with its unicode character.
4. <kbd>TAB</kbd> does not replace LaTeX strings with unicode characters.

As with #82, I've tried to break this up into conceptually cohesive commits that can be reviewed independently.

Demonstration gif:

julia-emacs-pr-100-demo-2


non-Jedi:
@tpapp I was going to push an additional commit that also overrides <kbd>TAB</kbd> to dynamically do `expand-abbrev` as well as completion, but it results in some ambiguous situations: what should be the result of pressing <kbd>TAB</kbd> with `\sqrt` entered: `√` or `\sqrtbottom`? The REPL currently complets it to `√`, but I'm not convinced this is the best solution for emacs.

I would prefer to keep the concept of completion and abbrevs separate conceptually and possibly  bind some other easily accessible key combination to `expand-abbrev`. Thoughts?

tpapp:
Sorry for not getting back to you on this yet, I will need some time to review it but may not get to it until the middle of this week.

I will think about this question too.

julia-mode.el
```diff
@@ -787,6 +787,15 @@ Return nil if point is not in a function, otherwise point."
   (set (make-local-variable 'indent-line-function) 'julia-indent-line)
   (set (make-local-variable 'beginning-of-defun-function) #'julia-beginning-of-defun)
   (set (make-local-variable 'end-of-defun-function) #'julia-end-of-defun)
+  ;; If completion before point has higher priority than around, \lamb
+  ;; can get completed to \lambdamb
+  (add-hook 'completion-at-point-functions #'julia-mode-latexsub-completion-at-point-before)
+  (add-hook 'completion-at-point-functions #'julia-mode-latexsub-completion-at-point-around)
+  (when julia-force-tab-complete
+    (setq-local tab-always-indent 'complete)
```
non-Jedi: The macro `setq-local` was added in emacs 24.3. I'm assuming we're only supporting the latest minor version of all the major versions of emacs we support, but if not, this needs to be changed to `(set (make-local-variable 'tab-always-indent) 'complete)`. I can also add a commit changing all usage of `(set (make-local variable ...) ...)` to `setq-local` for stylistic consistency if desired.

tpapp: I am fine with this. Let's keep things simple and go with 24.3 as in #101.

tpapp:
Thanks for doing this, and in such a clean way. I like the code.

I would like to merge this in some form, but I am wondering what the best interface is (or if people prefer various versions, how to support them all) so I am inviting people to comment on how they use completion.

Personally, I would prefer the following: a **single** <kbd>TAB</kbd> 

1. **first tries to indent**. if it did, then it is done. if the indentation was unchanged, then:
2. whenever the point is in or immediately after a **complete** LaTeX abbreviation, it is **replaced** by the relevant abbrev, eg `\sqrt` is replaced by `√`,
3. if no complete abbreviation is found, the abbrev window pops up with possibilities (if there are any). so <kbd>TAB</kbd> on `\sqr` would offer `\sqrt`, `\sqrint`, and `\sqrtbottom`. If the users selects one, the replacement would just happen without further prompting.

The motivation for the above is that trying to enter a `\alpha` into a Julia buffer is probably not a frequent use case, the user just wants the `α`, so we should go for it immediately without further keypresses. I find having to <kbd>Space BackSpace</kbd> cumbersome.

I don't know enough about `abbrev` to say whether that is the right implementation for this, but I am fine with using it.

julia-mode.el
```diff
@@ -50,6 +50,13 @@
   :type 'integer
   :group 'julia)
 
+(defcustom julia-force-tab-complete t
+  "Use Tab for completion instead of M-Tab in `julia-mode'.
+This overrides `tab-always-indent' in `julia-mode' buffers. It also
+enables `abbrev-mode' so that inserting a space character will replace
+a LaTeX string with a unicode symbol."
+  :type 'boolean)
+
```
nverno: If you want to `expand-abbrevs` after any punctuation, eg. \\alpha* => {laxex  alpha}*, you could add 
a local `post-self-insert-hook` along the lines of

```
(defun julia-abbrev-expand-after-symbols-hook ()
  (and (memq (car (syntax-after (1- (point)))) '(0 3))
       (setq this-command 'julia-abbrev-expand-after-symbols)
       (expand-abbrev)))
```
which I use in, eg. elisp, where I want my abbrevs to expand after I enter a ')' for example.
It could be another user choice to consider.  IIRC I needed to add this as a `post-self-insert` command because abbrev expansion after word syntax was hard-coded in C, but the implementation may have changed.

julia-mode.el
nverno: One issue I've noticed, is now my current `julia-mode-abbrev-table` elements don't expand anymore due to the addition of the specialized prefix.  Perhaps the latex table could have a non-default name?

nverno:
I also really like the completion working in the middle of symbols, as it does in elisp.

nverno:
> I would prefer to keep the concept of completion and abbrevs separate conceptually and possibly > bind some other easily accessible key combination to expand-abbrev. Thoughts?

Agreed, I think intertwining them is likely to break stuff.  Abbrevs have been around forever, having their own distinct behaviour.  Also, as with everything, the more modular the better - you would want to allow people to opt out of abbrevs, but use completion, and vice versa IMHO.

julia-mode.el
non-Jedi: The problem with this is all of the LaTeX codes that include punctuation (or even parentheses). Some examples: `\^)`, `\_=`, `\:soccer:`, `\^+`. I guess in practice, as long as none of the LaTeX codes that include symbols have other LaTeX codes that are superstrings of them cut off at the symbol it could be fine.

julia-mode.el
non-Jedi: I think the `julia-mode-abbrev-table` symbol rightfully belongs to this package, and you should expect to sometimes have symbols overridden if you put them in a package's namespace.

non-Jedi:
>I don't know enough about abbrev to say whether that is the right implementation for this, but I am fine with using it.

It would be a bit hacky, but I can build that interface on top of capf and abbrev, which would still allow people to customize a variable to use capf and abbrev in the emacs default manner. As mentioned above, doing so would not be my preference, but I can certainly implement that if it's the consensus.

non-Jedi:
>The only thing I would change is giving the latex abbrev table a non-default name, so that user defined abbrevs still work (eg. they aren't also required to have the special expansion regexp requirement). An alternative could be to add the latex abbrev table as a parent of the julia-mode-abbrev-table.

Could you expand on what you mean by here? Specifically "default name" isn't clear to me, nor is it clear how this interferes with regexps for user-defined abbrevs. And what latex abbrev table are you recommending we inherit from?

julia-mode.el
tpapp: > I don't think converting \sqrt to √ without further input is a good solution, cause you might still be planning to write something else

Then this may just be my personal preference --- if I want to choose between `\sqrt` and `\sqrtbottom`, I would complete on `\sqr` and pick from the popup. If feasible, keeping this modular and customizable would be nice.

julia-mode.el
nverno: I does, but you've given it a prefix that won't let any normal abbrevs work

nverno:
Call it `julia-mode-latex-abbrev-table` and set it as the parent of `julia-mode-latex-table` -- then you get normal behaviour

julia-mode.el
nverno: >  Some examples: \^), \_=, \:soccer:, \^+

Good point -- the above is too simple for this case

non-Jedi:
I've rebased on master and pushed a commit fixing the issue with the regex clobbering user abbrevs noted by @nverno. Now just awaiting feedback/decision on integrating LaTeX substitution/expansion.

tpapp:
Can you briefly summarize that the alternatives are, and the trade-offs (if any?)

non-Jedi:
Sure. Trying to be as brief as possible:

1. Leave as is in this PR.
  - Unless user customizes `julia-force-tab-complete`, <kbd>TAB</kbd> tries to indent then complete to LaTeX code, and either inserting a whitespace char or calling <kbd><kbd>M</kbd>-<kbd>x</kbd></kbd> `expand-abbrev` (bound to <kbd><kbd><kbd>C</kbd>-<kbd>x</kbd></kbd> <kbd>a</kbd> <kbd>e</kbd></kbd>) causes the unicode character to replace the LaTeX code.
  - Advantage: Fits naturally into the builtin Emacs concepts of completion (substring to full string) and abbrevs (string to other string) which will be less surprising for people used to Emacs conventions for other languages.
  - Advantage: less code to maintain and generally the most robust solution.
  - Disadvantage: Requires more key-strokes.
  - Disadvantage: Inserting LaTeX symbols into a string, docstring, or comment becomes difficult.
  - Disadvantage: `abbrev-mode` and `completion-at-point` need identical tables in slightly different data structures, so twice as much memory is used.
2. Add an `:exit-function` which calls `expand-abbrev` to the `completion-at-point` functions to replace LaTeX codes with unicode symbols as was done in #82. We can check a customizable variable to decide whether to do the replacement or not. To make this work robustly with the many completion frameworks out there, the `:exit-function` will have to run on both `'sole` and `'finished`.
  - Advantage: feels very natural to use--similar to Julia LaTeX substitution in the REPL (with `company-mode`).
  - Disadvantage: it's unclear when a properly implemented completion framework might call `:exit-function` with `'sole` and how performing the substitution before the completion is `'finished` might screw things up.
  - Disadvantage: `abbrev-mode` and `completion-at-point` need identical tables in slightly different data structures, so twice as much memory is used.
3. Do option 2 but only when `:exit-function` called with `'finished`.
  - Disadvantage: `ivy` is buggy (abo-abo/swiper#2345), so users using `ivy` won't have their LaTeX codes substituted.
  - Advantage compared to #82: Those users can still manually call `expand-abbrev`.
4. Do option 1 but bind <kbd><kbd>M</kbd>-<kbd>s</kbd></kbd> to `expand-abbrev`.
  - Less keystrokes than 1 but still more than 2.
5. Create our own function to bind to <kbd>TAB</kbd> which calls out to `completion-at-point` and manually substitutes unicode characters afterward.
  - Disadvantage: `company-mode` doesn't use `completion-at-point` to trigger it's completion, so this approach is incompatible with it.
  - Disadvantage: requires reimplementing a lot of Emacs machinery that already exists.
6. Turn all of this into a [company backend](https://github.com/company-mode/company-mode/wiki/Writing-backends) instead.
  - Advantage: company gives significantly better completion experience than Emacs out of the box.
  - Advantage: probably allows us to implement both completion and substitution at once without making tradeoffs.
  - Disadvantage: Probably jarring for somebody used to builtin completion, ivy, or auto-complete.
  - Disadvantage: julia-mode is no longer dependency-free.

There may be other options and advantages/disadvantages I've missed, but those are the ones that have been bouncing around in my head.

tpapp:
Sorry for being dense, but I am somewhat confused. I have re-read the history, and found that the original motivation of this string of PRs, starting with #80, was that

> This would integrate nicely with e.g. company and would generally make this package a better emacs citizen.

I am generally in favor of doing these things, but from various experiments it appears to be difficult to do this in a way that preserves the single-keystroke behavior we have currently. Given that we have a working solution, should we switch to one of the backends at this cost, just to be a better Emacs citizen?

Note that I like the current setup, so it is possible that I am missing something that someone else's workflow/setup requires. I am happy to accommodate that, but I am a bit lost why this is needed. 

That said, this PR has a lot of code improvements which should be incorporated.

non-Jedi:
I'm afraid my phrasing in #80 reflects my poor understanding at the time of how completion works in Emacs. Sorry about the confusion caused. The real benefit here is that by tying into `completion-at-point`, we get real **completion** of LaTeX strings (`\lam` to `\lambda`) and substitution rather than just substitution of already-complete LaTeX strings (`\lambda` to `λ`).

non-Jedi:
@tpapp what's the path forward here given the deafening silence of user-input on desired behavior?

tpapp:
Apologies for the lack of feedback. This is something I want to think about a bit and experiment with, but I am overwhelmed with other stuff at the moment. I understand that putting in the work and not getting it merged can be frustrating, but I ask for your further patience.

non-Jedi:
No problem. And no pressure from me with respect to getting it or any pull-request i make merged. I am already happily using all changes I've proposed on my own machines. Just wanted to make sure you weren't waiting on some other input or change from me. :)

non-Jedi:
@tpapp you had time to look at this yet? I understand if not, but I thought I'd ask since rebasing my personal `devel` branch onto this when I make an unrelated PR is becoming slightly burdensome.

I thought I should also explain a bit better my objection to our current approach: it breaks composibility (and doesn't offer real completion). Emacs has builtin facilities for both "completion" (`\la` to `\lambda` is done by `completion-at-point`) and "replacement" (`\lambda` to `λ` is done by `abbrev-mode`). Both have all sorts of hooks that users and other emacs packages can use to customize how they work; if we implement our own version of either (as `julia-latexsub` currently does for replacement), none of that works. It's the same reason it's generally a good idea to use the "Iterable" interface from `Base` rather than rolling your own: you get all sorts of great interop and functionality for free.

rfourquet:
I should have read this last comment first instead of reading in order the issue, your previous PR and this one, I would have understood faster :D 
I'm quite unfamiliar with all this stuff, so I don't have much useful feedback right now, but will use this code for the next few days to have a feel of what it does, and hopefully can then have a more informed opinion.

One thing I don't really understand yet is why <kbd>TAB</kbd> is used instead of (or in addition of) <kbd>M-TAB</kbd> for `completion-at-point`, as is done in e.g. `emacs-lisp-mode` (I know that <kbd>M-TAB</kbd> is not convenient on systems which also use this to switch windows... but <kbd>C-M-i</kbd> seems to be an alternative), and why it's better to have <kbd>TAB</kbd> bound to `completion-at-point` rather than to `abbrev-expand`. Along the read, I came to believe that one goal was to not redefine <kbd>TAB</kbd>.

non-Jedi:
Do note that the current implementation doesn't **rebind** TAB but merely sets the customizable variable `tab-always-indent` to `'complete`:

>If ‘complete’, TAB first tries to indent the current line, and if the line
>was already indented, then try to complete the thing at point.

To be honest, I would personally prefer not mucking around with this at all (I implemented it based on the opinion from Tamas and Mosè), but it is much less invasive than the current approach which entirely rebinds TAB. It also gives `julia-mode` similar behavior to other julia editors/REPLs in that you can tell somebody to press tab after a `\\` to complete latex symbols, and it pretty much universally works.

rfourquet:
> Do note that the implementation in this PR doesn't rebind TAB but merely sets the customizable variable tab-always-indent to 'complete:

Ah right sorry. I even read about this variable yesterday, but got saturated with information and got no time anymore to look deeper in the patch.

> To be honest, I would personally prefer not mucking around with this at all (I implemented it based on the opinion from Tamas and Mosè), but it is much less invasive than the current approach which entirely rebinds TAB

I fully agree on both points. In the future we can discuss whether we want to clean it up a bit further (e.g. by letting the user set `tab-always-indent` herself).

> It also gives julia-mode similar behavior to other julia editors/REPLs in that you can tell somebody to press tab after a \ to complete latex symbols, and it pretty much universally works.

Sounds good indeed, that might justify being slightly "opinionated".

> I'll go ahead and rebase this on current master so that it's easier for people to try out.

Excellent, thank you very much for all this work! @tpapp @nverno @giordano : if you are not doing it already, would you mind using this branch as your daily driver for a short while so that we give a fair chance to this patch and get a feel for the "ergonomic" change involved?
(you can do `git fetch origin pull/100/head:p100; git checkout p100` if your upstream remote is named `origin`).

tpapp:
@rfourquet: thanks for the reminder. @non-Jedi: apologies for still not reviewing this, I will be using this branch from now and and should be able to give feedback within a few days.

tpapp:
I had a change to take it for a test ride. I want to confirm that I am using it the right way, and nothing in my setup interferes. 

1. I type `\alph|` and press <kbd>TAB</kbd> (as above, `|` is the point).
2. Since I am using helm, a `*helm-mode-completion-at-point*` buffer pops up. In that, `\Alpha` comes before `\alpha`, which is a minor inconvenience.
3. I pick the `\alpha α` line, then I get `\alpha |` in the buffer (note space).
4. then to get `α`, I *delete the space* and then insert a whitespace char.

Is this how it should work, or is there a quirk in my setup that interferes somehow? TBH I find this somewhat inconvenient.

non-Jedi:
I've recorded a 30-second screencast showing how this PR works in vanilla emacs with both vanilla completion and company-mode and edited it into the OP.

tpapp:
Thanks for the screencast. It must be a quirk with my local setup; I will try to investigate.

Is there a way of retaining the current behavior for users that prefer it, eg by customizing some variables? Ie go straight from `\alph` to `α` without going through `\alpha `, which I rarely ever need when coding Julia.

non-Jedi:
I can make that change if that's the direction decided for this functionality. It will require one of the tradeoffs mentioned above (https://github.com/JuliaEditorSupport/julia-emacs/pull/100#issuecomment-601785146) with respect to `:exit-function` due to inconsistent support in existing completion frameworks.

tpapp:
Thanks for the reminder, and apologies for not remembering that nice summary about the trade-offs.

I see little use case for completing into LaTeX symbols like `\alpha` *per se* --- the end goal for Julia users is Unicode input. Because of this, I am somewhat unwilling to give up the current interface (just press <kbd>TAB</kbd> and things work) for something that requires more keystrokes, even if it is conceptually nicer.

You suggested that solutions relying on `:exit-function` are buggy so I guess those are out.

If I understand correctly, parts of this PR implement completing for unfinished symbols (eg `\la|m`). Can those could be just backported to the current framework without a change in the user interface? That would be very nice.

Then given that we have a list of completions, users who want some integrated Emacs completion mechanism could use it with something like https://github.com/vspinu/company-math, ideally in package depending on the relevant package and this one (which contains the list). This is option 6 in the tradeoffs writeup and seems like the best solution.

rfourquet:
> I see little use case for completing into LaTeX symbols like \alpha per se --- the end goal for Julia users is Unicode input.

Note that this is also the REPL behavior; I occasionally missed the possibility in emacs to latex-complete a symbol whose full name I didn't know. So at least when there is ambiguity, I find it useful to see all the possible completions rathen than automatically complete into the shortest one.

> Because of this, I am somewhat unwilling to give up the current interface (just press TAB and things work) for something that requires more keystrokes, even if it is conceptually nicer.

Given that there is currently no completion at all for latex symbol, I would see this PR as a strict improvement: e.g. type `\alp<TAB>+3` and you get `α+3`; currently you have to do `\alpha<TAB>+3`, which is more keystrokes. (Sorry if I'm missing something).

non-Jedi:
> You suggested that solutions relying on `:exit-function` are buggy so I guess those are out.

I'm going to try option 2 with :exit-function from above <https://github.com/JuliaEditorSupport/julia-emacs/pull/100#issuecomment-601785146>. It might be less of a problem in practice than I'm expecting.

> If I understand correctly, parts of this PR implement completing for unfinished symbols (eg `\la|m`). Can those could be just backported to the current framework without a change in the user interface? That would be very nice.
>
> Then given that we have a list of completions, users who want some integrated Emacs completion mechanism could use it with something like https://github.com/vspinu/company-math, ideally in package depending on the relevant package and this one (which contains the list). This is option 6 in the tradeoffs writeup and seems like the best solution.

Am I understanding correctly this idea is that this PR adds the functions to `completion-at-point-functions` but does not change TAB behavior from current? So pressing TAB would not do completion (unless the user had manually customized `tab-always-indent`), but it would substitute for `\lambda|` as it does currently?

tpapp:
@rfourquet:

> Note that this is also the REPL behavior

Almost --- in the REPL, you can just keep pressing <kbd>TAB</kbd>. I would be perfectly happy to do that in `julia-mode` too, an extra <kbd>TAB</kbd> should not matter much. 

That said, I think this is a convincing argument in favor of the API change in this PR, as it introduces consistency between the REPL and the code buffers.

@non-Jedi:

> Am I understanding correctly this idea is that this PR adds the functions to completion-at-point-functions but does not change TAB behavior from current? So pressing TAB would not do completion (unless the user had manually customized tab-always-indent), but it would substitute for \lambda| as it does currently?

I think that completion is a great feature and a really nice contribution of this PR, I just want a streamlined experience for users who prefer the current behavior (minor modifications are of course OK). 

Would it be possible to make <kbd>TAB</kbd> *also* expand abbrevs if applicable? Eg

1. `\alpha|` <kbd>TAB</kbd> would replace to `α`
2. `\alp|` <kbd>TAB TAB</kbd> would replace to `α` after expansion

BTW, I narrowed the extra space issue above to be helm-specific, now I am debugging further, but that may be a quirk in my setup so it will not block merging this.

non-Jedi:
I've pushed an additional commit that implements option 3 from <https://github.com/JuliaEditorSupport/julia-emacs/pull/100#issuecomment-601785146>. The more I think about it, the more this option seems like a good compromise since `ivy-mode` users can still either enable `abbrev-mode` or call `expand-abbrev`. I also think that `ivy-mode` for in-buffer completion is relatively unpopular compared to other options.

Please try it out and let me know what you think.

tpapp:
I have been testing this, and so far it works fine and does exactly what I want. @non-Jedi, I appreciate all the fine work. I will get back here middle of next week after I used it for actual coding for a while, then we should merge.

As usual, code reviews would be appreciated, I will do one too.

tpapp:
For some reason now helm-mode goes a bit crazy with the latest updates, trying to complete `\bet` to `\eta`. @non-Jedi, can you check this on your system?

If you can't replicate, don't bother investigating, I think that I will switch to company mode for Julia anyway. I have years of gradually accumulated cruft in my helm setup which I should wade through at some point, but again it should not block this PR.

julia-mode-tests.el
```diff
@@ -751,27 +751,70 @@ return fact(x)
...
-  (should (equal (julia--substitute "\\alpha " 7) "α "))
-  (should (equal (julia--substitute "x\\alpha " 8) "xα "))
-  (should (equal (julia--substitute "\\kappa\\alpha(" 13) "\\kappaα("))
-  (should (equal (julia--substitute "\\alpha" 7) "α"))
-  ; (should (equal (julia--substitute "\\alpha" 6) "α")) ; BROKEN
+(ert-deftest julia--test-latex-abbrev ()
+  (should (equal (julia--abbrev "\\alpha " 7) "α "))
+  (should (equal (julia--abbrev "x\\alpha " 8)  "xα "))
+  (should (equal (julia--abbrev "\\kappa\\alpha(" 13)  "\\kappaα("))
+  ; (should (equal (julia--abbrev "\\alpha(" 6)  "α")) ; BROKEN
```
tpapp, line 793: Is this still broken?

julia-mode-tests.el
```diff
@@ -751,27 +751,70 @@ return fact(x)
...
   )
 
+(defun julia--call-latexsub-exit-function (contents beg position name auto-abbrev)
+  "Return buffer produced by `julia--latexsub-exit-function'."
+  (with-temp-buffer
+    (insert contents)
+    (goto-char position)
+    (setq-local julia-automatic-latexsub auto-abbrev)
+    (funcall (julia--latexsub-exit-function beg ) name 'finished)
```
tpapp, line 802:
```suggestion
    (funcall (julia--latexsub-exit-function beg) name 'finished)
```
(space)

tpapp:
I am ready to merge this, so pinging the usual suspects: @giordano @FelipeLema @rfourquet, reviewing this would help if you have the time.

julia-mode-tests.el
non-Jedi: Yes. Haven't thought of a good, performant way of fixing mid-symbol LaTeX completion followed immediately by non-whitespace.

non-Jedi:
> For some reason now helm-mode goes a bit crazy with the latest updates, trying to complete `\bet` to `\eta`. @non-Jedi, can you check this on your system?

I can reproduce this. Additionally, after updating helm, I'm seeing the issue you described above where a space is inserted after the completion (which prevents the `:exit-function` from working as intended since `point` isn't where it expects it to be. So there appear to be 2 upstream problems with helm:

1. `\bet` bizarrely interpreted as `\et` when doing completion.
2. Space being inserted by helm before `:exit-function` is run.

When I get a chance, I'll try to create a minimum reproducer to file an issue against helm.

tpapp:
Completion does not work for me if the point is before a `)`. Eg on `(\eta|)`, with `|` representing the point. Can you reproduce this?

non-Jedi:
> Completion does not work for me if the point is before a `)`. Eg on `(\alpha|)`, with `|` representing the point.

I can reproduce (with company and vanilla completion--not just helm). Will take a look. Thanks for finding the bug.

non-Jedi:
> Completion does not work for me if the point is before a `)`. Eg on `(\alpha|)`, with `|` representing the point. Can you reproduce this? (edit: attaching backtrace)

Turns out this is a known bug in emacs. Comment from `completion--capf-wrapper`:

>;; FIXME: Here we'd need to decide whether there are
>;; valid completions against the current text.  But this depends
>;; on the actual completion UI (e.g. with the default completion
>;; it depends on completion-style) ;-(
>;; We approximate this result by checking whether prefix
>;; completion might work, which means that non-prefix completion
>;; will not work (or not right) for completion functions that
>;; are non-exclusive.

So the long and short of it is that we can either have completion
within LaTeX symbols (`\lamb|da ` works but `\lambda|(` doesn't), or
we can have completion at end of LaTeX symbols (`\lambda|(` works but
`\lamb|da` doesn't); `\lambda| ` works fine with either option.

> ```
> Debugger entered--Lisp error: (wrong-type-argument number-or-marker-p nil)
>   abbrev-insert(nil #("\\alpha)" 0 1 (fontified t syntax-table (1)) 1 6 (fontified t) 6 7 (fontified t face (rainbow-delimiters-depth-1-face))) 2 9)
>   #f(compiled-function (name status) #<bytecode 0x1e5fdcd>)(#("\\alpha)" 0 1 (fontified t syntax-table (1)) 1 6 (fontified t) 6 7 (fontified t face (rainbow-delimiters-depth-1-face))) finished)
>   completion--done(#("\\alpha)" 0 1 (fontified t syntax-table (1)) 1 6 (fontified t) 6 7 (fontified t face (rainbow-delimiters-depth-1-face))) finished "Sole completion")
>   completion--do-completion(#<marker at 2 in tmp.jl> 9)
>   completion--in-region-1(#<marker at 2 in tmp.jl> 9)
>   #f(compiled-function (start end collection predicate) #<bytecode 0x1cc12f>)(#<marker at 2 in tmp.jl> 9 #<hash-table equal 3324/3694 0x170df0d> nil)
>   apply(#f(compiled-function (start end collection predicate) #<bytecode 0x1cc12f>) (#<marker at 2 in tmp.jl> 9 #<hash-table equal 3324/3694 0x170df0d> nil))
>   #f(compiled-function (funs global args) #<bytecode 0x1e5fe05>)(nil nil (#<marker at 2 in tmp.jl> 9 #<hash-table equal 3324/3694 0x170df0d> nil))
>   completion--in-region(#<marker at 2 in tmp.jl> 9 #<hash-table equal 3324/3694 0x170df0d> nil)
>   completion-in-region(#<marker at 2 in tmp.jl> 9 #<hash-table equal 3324/3694 0x170df0d> nil)
>   completion-at-point()
>   indent-for-tab-command(nil)
>   funcall-interactively(indent-for-tab-command nil)
>   call-interactively(indent-for-tab-command nil nil)
>   command-execute(indent-for-tab-command)
> ```

I should make this code more resilient so that if abbrev expansion
doesn't work, it merely fails instead of throwing--regardless of
what's decided above.

tpapp:
Just to step back a bit: my understanding is that the original motivation for this PR was

1. better integration with completion/replacement facilities of Emacs,
2. completion from inside symbols

From the above, it seems that 2. will not work, and 1. may require a lot of workarounds and offer incomplete functionality with some frameworks, because the completion frameworks/implementations are not mature enough. Is this correct?

While generally duplicating functionality is usually a waste of resources, I am wondering if it would be worthwhile to keep our current completion mechanism around for a while and revisit this issue later, maybe in a year.

@non-Jedi, what do you think? I recognize that you have invested a lot of work into this.

non-Jedi:
I was very frustrated to discover the aforementioned bug. Give me a couple days to cool down, and I'll see what I'm thinking.

tpapp:
Sure, take your time. There is no rush. Again, your work is very much appreciated and I am sorry to see this PR run into one bug after another in other software.

In any case, this PR has code improvements that should not be lost even if we stick to our own completion mechanism.

rfourquet:
(Just to apologies to not have followed in the last week or so, am on holidays now, will try to catch up when I can, and will be happy to do a code review when needed)

make-julia-latexsubs.jl
```diff
@@ -12,27 +12,40 @@ import REPL.REPLCompletions: latex_symbols, emoji_symbols
...
+function write_latexsubs_tables(io, src, hashvar, abbrevvar)
     for (k, v) in sort!(collect(src), by = last)
         ks = escape_string(k)
         vs = escape_string(v)
         if occursin(r"^\\U[0-9A-Fa-f]+$", vs)
             # codepoints outside the BMP can be problematic in older Emacsen
             cp = vs[3:end]
-            println(dest, "    (let ((c (decode-char 'ucs #x$cp)))\n",
-                    "      (if c (puthash \"$ks\" (char-to-string c) $(varname))))")
+            println(io, "    (let ((c (decode-char 'ucs #x$cp)))")
```
FelipeLema, line 22: While it may sound redundant to do `println` for each line, I love it how clean and easy-to-read this block ended up begin

non-Jedi:
Thanks for taking the time to review @FelipeLema and @tpapp. Reviewing PRs tends to be one of the most drudgy and thankless tasks in OSS; I really appreciate y'all taking the time to do it.

>From the above, it seems that 2. will not work, and 1. may require a lot of workarounds and offer incomplete functionality with some frameworks, because the completion frameworks/implementations are not mature enough. Is this correct?

I would add a third point:

3. Offer completion of LaTeX symbols instead of just replacement of already-complete symbols.

Remember that currently only `\lambda` can be replaced with `λ`. This PR allows completion of `\l` as well. To me, this is the most important part of this PR, and it remains intact.

For point 1, Thierry has been crazy helpful on the Helm side, so things should mostly work as well there as with emacs native completion (I have one more bug I need to file). This leaves ivy as the only completion framework I'm aware of which doesn't work well with `:exit-function` (there's also selectrum which is a new framework this year, but I haven't tested it yet). Workarounds already exist in this PR for any completion framework that doesn't call `:exit-function` quite correctly in that users can manually enable `abbrev-mode`.

For point 2, it isn't a regression from current julia-mode abilities, and once this PR is merged, I may try to sneak something in to `julia-indent-function` to add inside-symbol substitution. The important point is that this change doesn't preclude us from adding inside-symbol substitution in the future.

Given the above, I think the path forward is to push a commit that removes the code triggering the `FIXME` bug in minibuffer.el. If you agree @tpapp, I'll go ahead and do so.

tecosaur:
Hey all! I may have missed something, but having read a bit of this I have a question.

Early on @tpapp says:

> The motivation for the above is that trying to enter a `\alpha` into a Julia buffer is probably not a frequent use case, the user just wants the `α`, so we should go for it immediately without further keypresses. I find having to Space BackSpace cumbersome.

I agree, and I'm looking here in the first place because I don't know why I have to press `TAB` so much. If I type `\alpha\^2\pm\epsilon` I don't know what I'd want other than `α²±ϵ` so I don't think I should need to `TAB` four times. I find this particularly annoying with accents and subscripts.

As such, I'm wondering if Julia coding would benefit from a TeX-like input method, like the [TeX input method](https://github.com/emacs-mirror/emacs/blob/master/lisp/leim/quail/latin-ltx.el) but just with Julia-supported symbols. If you haven't used input methods before, see a clip of me using the TeX input method below:

tex-input-method

I'd need fourteen `TAB`s in the above if just using Julia mode.

tpapp:
@non-Jedi, apologies for not getting back to you on the above yet. I reread your [summary above](https://github.com/JuliaEditorSupport/julia-emacs/pull/100#issuecomment-601785146), but I have to admit that I kind of lost track about where this PR stands in terms of feasibility and advantages. 

I think we would benefit from separating *interface* and *implementation* questions to the extent possible. It is very clear from the discussion that people expect different completion behavior and we should do our best to accommodate it with offering customizable options.

Whether that should be implemented with our own framework or hooking into one of the existing ones is a question I don't have a good answer to because I am less familiar with the details and corner cases. Ideally, avoiding code duplication is always best, but I am wondering if it is practical in this case or whether we should stick to our own, existing solution and keep tweaking it to suit users.

julia-mode-latexsubs.el
```diff
@@ -1,3327 +1,6660 @@
...
 ;; This file is automatically generated. Do not edit by hand.
+
+(define-abbrev-table 'julia-latexsub-abbrev-table ()
+  "Abbrev table for Julia LaTeX substitution."
+  :regexp "\\(\\\\[[:graph:]]+\\)")
+
 (defconst julia-mode-latexsubs
   (let ((table (make-hash-table :test 'equal)))
     (puthash "\\exclamdown" "¡" table)
+    (define-abbrev julia-latexsub-abbrev-table "\\exclamdown" "¡" nil :system t)
```
tecosaur, line 11: I'm curious, is there any particular reason why each hash entry is duplicated for the abbrev table instead of leaving the hash table alone and then looping through the entries to create the abbrev table at the end?

non-Jedi:
> @non-Jedi, apologies for not getting back to you on the above yet. I reread your [summary above](https://github.com/JuliaEditorSupport/julia-emacs/pull/100#issuecomment-601785146), but I have to admit that I kind of lost track about where this PR stands in terms of feasibility and advantages.

It's been a long, winding journey. I've lost track of it a bit myself. I think I need to do
a bit of work to make sure that this works cleanly with all the major completion frameworks,
so it's not ready to merge in its current state. The intention is that this PR will strictly
expand on `julia-mode`'s capabilities (everything that currently works with regards to
replacement still will). To be honest, I probably won't have time to polish this any time in
the next month or two; I'm moving states and starting a new job.

> I think we would benefit from separating _interface_ and _implementation_ questions to the extent possible. It is very clear from the discussion that people expect different completion behavior and we should do our best to accommodate it with offering customizable options.

`completion-at-point` already separates those two things rather cleanly. By hooking into it,
we can let users choose from the wide world of emacs completion frontends and behaviors.

> Whether that should be implemented with our own framework or hooking into one of the existing ones is a question I don't have a good answer to because I am less familiar with the details and corner cases. Ideally, avoiding code duplication is always best, but I am wondering if it is practical in this case or whether we should stick to our own, existing solution and keep tweaking it to suit users.

I really think **completion** of LaTeX codes (not just substitution) is an important
feature, so adding that is important. It would be a lot of work I'm not inerested in to
write our own completion framework.

non-Jedi:
Revisiting again, and I think this should be merged as is (once conflicts are resolved). A branch with these commits is what I've been using for the past two years, and having real completion for tex symbols has greatly improved my julia editing experience. Also, I can no longer trigger the error noted above when completing `(\alpha|)` on recent versions of emacs (completion still doesn't trigger, but no error is thrown). Just to clearly state the pros and cons as I see them of merging this:

## Cons
1. Code in these commits is less battle-tested and therefore likely to be buggy in ways that our current replacement setup is not.
2. Pressing <kbd>TAB</kbd> when a symbol immediately follows your completion candidate will not cause the candidate to be replaced. To work around this, the user has to either trigger `expand-abbrev` or insert another space or symbol.
3. That's the list. Other than those two points, this commit strictly improves over the previous behavior.

## Pros
1. Real completion when pressing <kbd>TAB</kbd> (can simply type `\a` <kbd>TAB</kbd> and then use your completion UI to select your final result instead of having to type `\alpha` <kbd>TAB</kbd>.
2. Many situations where you previously would have had to type <kbd>TAB</kbd>, and now it's no longer necessary.
    - e.g. typing `\alpha = ...` does not require pressing <kbd>TAB</kbd> at all to get the desired behavior.
3. No longer hijacking the keymapping of <kbd>TAB</kbd>, so any user customization that's been done remains intact.
4. Can replace tex symbols when `point` is in middle instead of only at end.

## Alternatives

A system implemented as an input method would let us solve #176 as well, but current input method packages don't have a nice way of tying into `completion-at-point`, and implementing one is a bit more work than I can commit to at the moment.

tpapp:
I am open to merging this as is; AFAICT there are no conflicts to fix (I think the recent commits took care of that). I will experiment with the branch myself a bit to see how it works. 

I would appreciate if others did the same, so this could get more testing.

tpapp:
Let's revisit merging in two weeks, on Dec 21. If I happen to forget, friendly pings are welcome.

julia-mode-latexsubs.el
non-Jedi: There isn't. I just found this the simplest way of writing it.

tpapp:
@non-Jedi: would you please share the additional setup and hooks you use for completion?  I have been using `company-mode` and it is a bit clunky with the new PR, but I am happy to experiment with something else. Just want to free-ride on someone else who figured this out.

non-Jedi:
> @non-Jedi: would you please share the additional setup and hooks you use for completion? I have been using `company-mode` and it is a bit clunky with the new PR, but I am happy to experiment with something else. Just want to free-ride on someone else who figured this out.

The entirety of my config as it relates to `company` is below, but there's nothing specific to Julia in there.

```elisp
(straight-use-package 'company)
(add-hook 'after-init-hook 'global-company-mode)

(with-eval-after-load 'company
  (define-key company-mode-map [remap indent-for-tab-command]
    'company-indent-or-complete-common)
  (define-key company-active-map (kbd "TAB") 'company-complete-common-or-cycle)
  (define-key company-active-map (kbd "<tab>") 'company-complete-common-or-cycle))
```

I do have the following in my config so that I can substitute unicode for latex code without going through completion by inserting a character with whitespace or punctuation syntax. That makes the example of doing `\alpha\^2\pm\epsilon` without pressing <kbd>TAB</kbd> work.

```elisp
(add-hook 'julia-mode-hook #'(lambda () (abbrev-mode 1)))
```

Could you share what specifically you find clunky about the combination of this PR and `company`?

tpapp:
In my current setup (which is a spaghetti of accreted stuff in my Emacs config and probably needs cleanup),

1. sometimes <kbd>tab</kbd> completes, sometimes it does not and I have to wait for the popup,
2. sometimes the instead of the popup I get a completion-list buffer above the minibuffer,
3. sometimes I cannot complete words after I typed them and go back, eg typed `\tau `, pressed left, and then no popup.

The problem is that what is happening is not deterministic. I am sure it has to do with my config. I will probably have to bisect it but your config will be a good starting point.

non-Jedi:
Ah. I think I see the problem. This PR makes `julia-mode` set the `tab-always-indent` variable to `'complete` so that `indent-for-tab-command` which is default mapped to <kbd>TAB</kbd> performs completion, but company doesn't go through `indent-for-tab-command` since it's whole thing is "auto" completion (without requiring <kbd>TAB</kbd>). So when you press <kbd>TAB</kbd> it tries to go through emacs native completion UI, but if you wait it goes through the company popup via its capf backend.

I don't think company provides a nice way for a mode to tell it to hook into `indent-for-tab-command`, but I'll think about it. In the meantime, you can [add the following to your emacs config so that pressing <kbd>TAB</kbd> triggers company completion](https://github.com/company-mode/company-mode/issues/94#issuecomment-365701801):

```elisp
(define-key company-mode-map [remap indent-for-tab-command]
    'company-indent-or-complete-common)
```

non-Jedi:
I've pushed a commit that should help this behave better with `company-mode`'s non-standard behavior. Should fix your problem, @tpapp.

tpapp:
Thanks. Can the warning
```
Compiling file /home/tamas/src/julia-emacs/julia-mode.el at Mon Dec 19 22:27:30 2022
Entering directory ‘/home/tamas/src/julia-emacs/’
Warning (bytecomp):  docstring wider than 80 characters
Warning (bytecomp): the function ‘julia-company-indent-for-tab-command’ is not known to be defined.
```
be somehow suppressed?

FWIW, I am always happy to drop support for old Emacs versions.

non-Jedi:
I'm not seeing byte-compilation errors locally after the last commit.

non-Jedi:
@tpapp Friendly ping. Are you ready to merge this? Should I push an additional commit changing minimum emacs version and test targets and a changelog update?

tpapp:
@non-Jedi: sorry I was sick for 2 weeks and I am still catching up.

I promise that I will merge this PR once I had a chance to evaluate it. The problem is that my emacs config contains a lot of idiosyncratic cruft that interferes with completion. I need to bisect and clean it up.

I realize thay you did a lot of work on this and it is frustrating to wait for one person. So if anyone else wants to merge this, feel free to do so

In the meantime, please understand that additional work you do on this will not be wasted and this will definitely get merged.

non-Jedi:
@tpapp I hope you're feeling better! It always stinks being sick over the holidays.

No pressure on getting this merged; just let me know if I can help any more with making this play nicely with your config. I appreciate you doing the thankless work of maintaining this and reviewing PRs.

tpapp:
@non-Jedi: thanks for your patience. I have cleaned up my Emacs config and it works fine (I think it was some interaction with some ido thing, but I did not track it down). I have now switched to corfu.

I have been testing it for a week and it is superb. I am now happy to merge this, I think it has been reviewed quite thoroughly so I will not ask for another round of reviews. I will wait until Monday for comments though.

tpapp:
**NOTE** after merging,

1. drop support for 25, require at least Emacs 26
2. remove 25.* from tests
3. bump version to 0.5, this is a significant feature change

tpapp:
I have an issue with this PR, I wonder if anyone can replicate: in the buffer I have

```julia
α\hat_mean
```

with the cursor on the `_`. I press <kbd>TAB</kbd>, menu pops up, I select `\hat`. Then I end up with

```julia
α̂
```

ie the `_mean` disappears.

tpapp:
@non-jedi: I have been studying your code, the Emacs completion system (of which I am completely ignorant), and thinking about the design of my ideal completion system.

It is clear that completion should always start on the `\`, and if we traverse backwards and encounter something that cannot be part of a valid symbol, we should cop out early. So I am fine with how `julia--latexsub-start-symbol` works.

I am wondering if more effort could be spent on finding the end though. Specifically, from the `\` search until the longest string is found which could be part of a valid prefix. Thus,
```ascii
\lam|da00
\lambda|=
```
would complete on `\lambda` (`|` is the point), while 
```ascii
\lam|=
\l|am_1
```
would complete on `\lam`. I realize that it may be expensive to do this, but perhaps an efficient data representation can be found for substrings, eg using a tree with valid next characters.

Then, given this, we could offer a pre-filtered table from the `julia-mode-latexsubs`, just with the valid completions based on the part we found. Does this make sense? Or do we have to return the whole list?

tecosaur:
Have you seen https://github.com/ymarco/auto-activating-snippets ?

tpapp:
@tecosaur: thanks, I was not aware of it. I would prefer to keep the completion depend only on the contents of the buffer and the location of the point.

tecosaur:
Sure, but it's doing exactly what you talked about regarding an efficient (tree) representation of valid next chars.

non-Jedi:
@tpapp the problem is that trying smartly decide where the entity you're completing from the middle of results in ambiguity.

Let's take `\l|am` as an example where `|` represents `point`. Our completion function needs to return a list of the form `(START END COLLECTION . PROPS)`. `start` is obviously 1 where the backslash is, but `end` could be either 3 or 4 depending on what the user wants. Maybe the user is trying to write `⪫m` (`\lat` `m`), and we should make `end` 3, or maybe the user is trying to write `λ` (`\lambda`). One of those seems much more likely to me as a human, but I don't know how to programmatically distinguish between the cases.

One hacky solution I could try would be to create a series of ~10 `completion-at-point-functions` with increasing static amounts of look-ahead (`(setq end (+ (point) lookahead)`, but doing so is more than a bit of a hack, and I'd start to be slightly worried about slowing  down completion in that case.

non-Jedi:
> I have an issue with this PR, I wonder if anyone can replicate: in the buffer I have
> 
> ```julia
> α\hat_mean
> ```
> 
> with the cursor on the `_`. I press TAB, menu pops up, I select `\hat`. Then I end up with
> 
> ```julia
> α̂
> ```
> 
> ie the `_mean` disappears.

I figured out what's going on here, but I'm not sure what the best way to fix it is. Here's
step-by-step how we end up in this situation:

1. `julia-mode-latexsub-completion-at-point-around` determines the completion candidates
   will be for the buffer positions corresponding to `\hat_mean`.
2. The `completion-at-point` machinery deterines that `\hat` is a potential completion for
   `\hat_mean` according to the user's `completion-styles`--specifically the `emacs22`
   completion style which ignores text in the buffer after `point`.
3. When the user selects `\hat` as the completion, `completion-at-point` does what it's
   supposed to do and replaces the completion of candidate of `\hat_mean` with the selected
   completion of `\hat`.
4. The `:exit-function` then replaces `\hat` with it's appropriate unicode character.

Here's roughly what we'd like to happen instead which is what does happen if `emacs22` is
removed from `completion-styles`:

1. `julia-mode-latexsub-completion-at-point-around` determines the completion candidates
   will be for the buffer positions corresponding to `\hat_mean`.
2. The `completion-at-point` machinery finds no potential completions for `\hat_mean`.
3. `completion-at-point` then calls `julia-mode-latexsub-completion-at-point-before` which
   determines that completion will be for the buffer positions corresponding to `\hat`.
4. From here, everything proceeds as the user would expect.

Right now, I only see a few options for fixing this:

1. Disable all completion styles other than `basic` in the `julia-mode` buffer.
   - This will subvert user expectations if they have any other
     `completion-at-point-functions` that are active in `julia-mode` buffers.
   - I don't think emacs provides any facilities for modifying `completion-styles` only for
     specific `completion-at-point-functions`.
2. Remove `julia-mode-completion-at-point-around`.
3. Pre-filter the selection of completions returned for
   `julia-mode-latexsub-completion-at-point-around`.
   - This is the option I'm currently planning on implementing.
   - The emacs manual advises against doing so, but I think our application makes sense as
     an exception to the guideline:

>Additionally, the COLLECTION should generally not be pre-filtered
>based on the current text between START and END, because that is
>the responsibility of the caller of ‘completion-at-point-functions’
>to do that according to the completion styles it decides to use.

tpapp:
At this point I have nothing against pre-filtering, but I realized that I cannot spare the effort of reading the whole [Completion](https://www.gnu.org/software/emacs/manual/html_node/elisp/Completion.html) chapter in the manual and then looking at what other language modes do. As typical for Emacs, there are a few hundred different ways of doing what we want, we should opt for the one that meshes the best with the completion frameworks people introduced recently. But I have no detailed knowledge of how these work.

Also, I am happy to get rid of "around" completion if we need to, or severely restrict it (eg only works on full escapes, or whatever we need).

tpapp:
The latest commit does not fix the disappearing suffix issue for me.

However, I am now warming to the idea of removing "around" completion. I previously thought that it is a nifty feature, but now I do not think it is so important. The main purpose of this feature is *input* of UTF-8 characters, ie you input them and you are done. Typing `\alpha`, not replacing, and then going back to get `α` is perhaps not something that we should handle via this mechanism, in Emacs it would be more natural to use `query-replace` etc for that.

But I would like to make this mode comfortable for all users if feasible, so please comment. @adam-higuera (cf #70)?

tpapp:
@non-Jedi: I brute-forced end lookup with a hash table, please take a look at https://github.com/non-Jedi/julia-emacs/pull/1, a PR against your branch. Personally I find it performant.

tpapp:
I have been testing it the above modification two days and find it nice, does exactly what I want. Is anyone else testing this PR in practice? Please try out https://github.com/non-Jedi/julia-emacs/pull/1 and let me know how it goes, especially wrt to performance and corner cases.

Now that tree-sitter is on the horizon I would like to merge this in a couple of weeks.

PR #101, opened 2020-03-02
Remove regexp-opt workarounds for Emacs < 24

non-Jedi:
Additional simplification enabled by #87.

This turned into a bit of a general cleanup PR with the subsequent commits.

### DIFF
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -1,9 +1,10 @@
 ;;; julia-mode.el --- Major mode for editing Julia source code -*- lexical-binding: t -*-
 
-;; Copyright (C) 2009-2014 Julia contributors
-;; URL: https://github.com/JuliaLang/julia
+;; Copyright (C) 2009-2014 Julia contributors, 2015-2020 julia-mode contributors
+;; URL: https://github.com/JuliaEditorSupport/julia-emacs
 ;; Version: 0.4
 ;; Keywords: languages
+;; Package-Requires: ((emacs "24.3"))
 
 ;;; Usage:
 ;; Put the following code in your .emacs, site-load.el, or other relevant file
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -65,11 +65,6 @@
 ;;;###autoload
 (add-to-list 'auto-mode-alist '("\\.jl\\'" . julia-mode))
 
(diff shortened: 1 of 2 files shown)


julia-mode.el
```diff
@@ -1,9 +1,10 @@
 ;;; julia-mode.el --- Major mode for editing Julia source code -*- lexical-binding: t -*-
 
-;; Copyright (C) 2009-2014 Julia contributors
-;; URL: https://github.com/JuliaLang/julia
+;; Copyright (C) 2009-2014 Julia contributors, 2015-2020 julia-mode contributors
+;; URL: https://github.com/JuliaEditorSupport/julia-emacs
 ;; Version: 0.4
 ;; Keywords: languages
+;; Package-Requires: ((emacs "24.3"))
```
tpapp: Do we actually need 24.3, or would 
```suggestion
;; Package-Requires: ((emacs "24"))
```
do?

julia-mode.el
```diff
@@ -906,10 +883,6 @@ following commands are defined:
 
 (add-hook 'inferior-julia-mode-hook 'inferior-julia--initialize)
 
-;;;###autoload
-(defalias 'run-julia #'inferior-julia
-  "Run an inferior instance of `julia' inside Emacs.")
-
```
tpapp: I guess that we can remove this, but technically it is a breaking change, so we should mention it in the changelog.

julia-mode.el
tpapp: Nevermind, I saw [this comment](https://github.com/JuliaEditorSupport/julia-emacs/pull/100/files#r386407333).

julia-mode.el
non-Jedi: I'm actually going to add this back in. I did a quick survey of other external language packages with a REPL, and the `run-LANGUAGE` function seems like a convention (Haskell had it, and I think ess had it).

non-Jedi:
I've force-pushed the last commit with `run-julia` added back in and the removal of `latexsub` added to the changelog.

julia-mode.el
tpapp: Thanks for looking into this.

PR #102, opened 2020-03-09
Font lock tweaks

### LINKED ISSUE #4: Improve highlighting of quoted symbols

non-Jedi:
Close #4 

See commit messages for justification for changes.

All of these changes only affect fontification (EDIT: with exception of last 2). Do I need a changelog entry?

### DIFF
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -243,7 +243,7 @@
    '("if" "else" "elseif" "while" "for" "begin" "end" "quote"
      "try" "catch" "return" "local" "function" "macro" "ccall"
      "finally" "break" "continue" "global" "where"
-     "module" "using" "import" "export" "const" "let" "do" "in"
+     "module" "using" "import" "export" "const" "let" "do"
      "baremodule"
      ;; "importall" ;; deprecated in 0.7
      ;; "immutable" "type" "bitstype" "abstract" "typealias" ;; removed in 1.0
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -57,7 +57,7 @@
   :group 'julia-mode)
 
 (defface julia-quoted-symbol-face
-  '((t :inherit font-lock-preprocessor-face))
+  '((t :inherit font-lock-constant-face))
   "Face for quoted Julia symbols, e.g. :foo."
   :group 'julia-mode)
 
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -197,7 +197,11 @@
 
 (defconst julia-forloop-in-regex
   "for +.*[^
-].* \\(in\\|∈\\)\\(\\s-\\|$\\)+")
+].* \\(in\\)\\(\\s-\\|$\\)+")
+
+(defconst julia--forloop-=-regex
+  "for +.*[^
+].*\\(=\\|∈\\)")
 
 (defconst julia-ternary-regex
   " +\\(\\?\\)[
@@ -279,6 +283,7 @@
     'font-lock-constant-face)
    (list julia-unquote-regex 2 'font-lock-constant-face)
    (list julia-forloop-in-regex 1 'font-lock-keyword-face)
+   (list julia--forloop-=-regex 1 'font-lock-keyword-face)
    (list julia-ternary-regex (list 1 'font-lock-keyword-face) (list 2 'font-lock-keyword-face))
    (list julia-function-regex 1 'font-lock-function-name-face)
    (list julia-function-assignment-regex 1 'font-lock-function-name-face)
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -267,6 +267,8 @@
       (group ":" (or letter (syntax symbol)) (0+ (or word (syntax symbol))))))
 
 (defconst julia-font-lock-keywords
+  ;; font-lock-builtin-face intentionally unused since any name from
+  ;; names(Base) can be aliased in a baremodule.
   (list
    ;; Ensure :: and <: aren't highlighted, so we don't confuse ::Foo with :foo.
    ;; (in Emacs, keywords don't overlap).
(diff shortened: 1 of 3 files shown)


julia-mode.el
```diff
@@ -243,7 +243,7 @@
    '("if" "else" "elseif" "while" "for" "begin" "end" "quote"
      "try" "catch" "return" "local" "function" "macro" "ccall"
      "finally" "break" "continue" "global" "where"
-     "module" "using" "import" "export" "const" "let" "do" "in"
```
tpapp: The commit message makes it clear why `in` was removed, and I agree, but would you mind adding a `julia--should-font-lock` test for `in` in `for ... in ...` expressions? And, possibly `for`, which we don't test for either.

tpapp:
LGTM. @giordano or @FelipeLema, can you please review?

tpapp:
> Do I need a changelog entry?

If possible, please make a brief one --- this does not break anything but it is an improvement.

tpapp:
I think that the commit comments

> In the future, we could consider using font-lock-builtin-face for
everything returned by names(Base), but given that none of these are
truly builtin (can be overridden in a baremodule), I think it's best
to not use this face.

> font-lock-type-face is described as being "for
the names of user-defined data types.". It is intended to font-lock
the point of type definition rather than use.

are very useful, and a short note about *not* using various font lock faces with a condensed form of this explanation would be useful to have as a comment near the relevant code, so that it is not added back by later PRs accidentally.

julia-mode.el
```diff
@@ -760,14 +734,13 @@ Return nil if point is not in a function, otherwise point."
...
-  (set (make-local-variable 'syntax-propertize-function)
-       julia-syntax-propertize-function)
-  (set (make-local-variable 'indent-line-function) 'julia-indent-line)
-  (set (make-local-variable 'beginning-of-defun-function) #'julia-beginning-of-defun)
-  (set (make-local-variable 'end-of-defun-function) #'julia-end-of-defun)
+  (setq-local comment-start "# ")
+  (setq-local comment-start-skip "#+\\s-*")
+  (setq-local font-lock-defaults '(julia-font-lock-keywords))
+  (setq-local syntax-propertize-function julia-syntax-propertize-function)
+  (setq-local indent-line-function 'julia-indent-line)
```
giordano: Shouldn't this be
```suggestion
  (setq-local indent-line-function #'julia-indent-line)
```
also for consistency with the lines below?

julia-mode.el
```diff
@@ -829,8 +802,7 @@ following commands are defined:
 \\{LaTeX-math-mode-map}"
       nil nil (list (cons (LaTeX-math-abbrev-prefix) LaTeX-math-keymap))
       (if julia-math-mode
-          (set (make-local-variable 'LaTeX-math-insert-function)
-               'julia-math-insert)))))
+          (setq-local LaTeX-math-insert-function 'julia-math-insert)))))
```
giordano: Also this one
```suggestion
          (setq-local LaTeX-math-insert-function #'julia-math-insert)))))
```

julia-mode.el
```diff
@@ -875,11 +847,11 @@ following commands are defined:
...
-  (setq comint-prompt-regexp julia-prompt-regexp)
-  (setq comint-prompt-read-only t)
-  (set (make-local-variable 'font-lock-defaults) '(julia-font-lock-keywords t))
-  (set (make-local-variable 'paragraph-start) julia-prompt-regexp)
-  (set (make-local-variable 'indent-line-function) 'julia-indent-line))
+  (setq-local comint-prompt-regexp julia-prompt-regexp)
+  (setq-local comint-prompt-read-only t)
+  (setq-local font-lock-defaults '(julia-font-lock-keywords t))
+  (setq-local paragraph-start julia-prompt-regexp)
+  (setq-local indent-line-function 'julia-indent-line))
```
giordano: And this one
```suggestion
  (setq-local indent-line-function #'julia-indent-line))
```

julia-mode.el
```diff
@@ -199,6 +199,12 @@
   "for +.*[^
 ].* \\(in\\|∈\\)\\(\\s-\\|$\\)+")
 
+(defconst julia-ternary-regex
+  " +\\(\\?\\)[
```
FelipeLema: I personally prefer using `rx` as it's easier to read, but leaving it as this looks ok.

FelipeLema:
other than my non-gating comment and comments from @giordano, looks ok to me

non-Jedi:
Thanks for the reviews everyone. I've pushed 2 commits addressing them (bdb667e, 91d37br). Apologies that bdb667e does more than adding a test, but I realized while writing the test that not all for-loops were handled.

I also pushed 77d66ab which simply removes a regex term for matching the `f{T}(x) = 1` parametric function syntax which was removed in Julia 1.0.

julia-mode.el
non-Jedi: I agree. By the time I realized it was hairy enough that `rx` would have improved readability, it was mostly complete and didn't seem worth rewriting :man_shrugging:. If you review bdb667e that I just pushed you'll find that I also wrote the regex manually there, but that one was intentional since it's very similar to the regex immediately above it, so it's simpler to see the differences if they're both written in the same form.

I'm happy to rewrite both using `rx` though if there's a strong consensus that it's worthwhile to do so.

tpapp: I am fine with it as is.

tpapp:
Thanks for all the reviews, and @non-Jedi again for this very nice PR and addressing all comments. I am going to merge this now.

PR #106, opened 2020-03-21
RFC: Make inferior-julia smarter

non-Jedi:
This is not at a point where's it's ready to merge yet.

Basically, I'd like feedback before I put more time into this on whether other people find this useful. Instead of having people using any one of [emacs-jupyter](https://github.com/dzop/emacs-jupyter), [ess](https://github.com/emacs-ess/ESS), or [julia-repl](https://github.com/tpapp/julia-repl) to get a Julia REPL in Emacs, why not just improve the one we already have?

Much of the work in this PR takes its inspiration from the builtin `inferior-octave`.

This PR implements:

- `?`, `]`, and `;` REPL modes (I wouldn't be against adding piecemeal support for things like the debugger or perhaps just creating an entirely separate comint-based mode for that).
- Navigate to the file/line shown in error stacktraces by clicking or hovering point and pressing enter.
- Rudimentary completion (see caveats in the second commit).

Future work:

- Use completion from `inferior-julia` in `julia-mode`.
- Allow multiple instances of `inferior-julia`.
- `eldoc-documentation-function` integration.
- etc.

I'm submitting this here not because I think it's ready to be merged (the issues in the completion commit need fixed first) but to get feedback on whether others would find value in me putting more time into this or whether everyone finds the existing situation with in-emacs Julia REPLs tolerable. Please try this PR and give any feedback. I'd hate to add a bunch of code that simply sits and rots if I ever lose use for it for whatever reason...

Thoughts? Do we want a full-featured REPL included in `julia-mode` or is this better relegated to external packages?

### DIFF
--- a/inferior-julia.el
+++ b/inferior-julia.el
@@ -29,7 +29,7 @@
   "Commandline arguments to pass to `inferior-julia-program'."
   :type '(repeat (string :tag "argument")))
 
-(defcustom inferior-julia-prompt-read-only comint-prompt-read-only
+(defcustom inferior-julia-prompt-read-only t
   "If non-nil, the Julia prompt is read only."
   :type 'boolean)
 
@@ -47,7 +47,7 @@
    "%s")
   "Format string to use with stacktraces with Julia Base paths.")
 
-(defvar inferior-julia-prompt-regexp "^\\w*> "
+(defvar inferior-julia-prompt-regexp "^julia> "
   "Regexp for matching `inferior-julia' prompt.")
 
 (defvar inferior-julia-error-regexp-alist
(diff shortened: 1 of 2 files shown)


tpapp:
>  finds the existing situation with in-emacs Julia REPLs tolerable

I am probably biased as I maintain julia-repl, but I find it sufficient for my needs (especially with vterm, which is [coming soon](https://github.com/tpapp/julia-repl/pull/84)). I am also enthusiastic @gcv's [julia-snail](https://github.com/gcv/julia-snail).

I would prefer if we kept this package focused on the mode for editing Julia code, mainly because I am not sure anyone is willing to commit to maintaining a REPL so I am concerned about bit rot.

If possible, I would suggest that experiments with inferior/REPL solutions happen in other packages.

sje30:
I'd say that if you have already done the work, go for it.  I wrote elsewhere about the multiple packages available for julia REPLs in emacs [1].   It seems like the support for Julia is in a state of flux, so perhaps it is worth a compare-and-contrast of several approaches?

[1] https://github.com/gcv/julia-snail/issues/1

non-Jedi:
>so I am concerned about bit rot.

Should the existing `inferior-julia` be removed then? I'm not sure *anybody* has been using it (for good reason; doing so is a chore), so it's just as prone to bit-rot as the improved version I'm proposing. It would be nice if I split this into a package if I could keep the `inferior-julia` prefix in order to ease possible future integration back into `julia-mode`...

I will note that not having a full-featured `inferior-julia` builtin to `julia-mode` makes us relatively unique among the REPL-oriented languages. The only other language I'm aware of that doesn't build an `inferior-LANG` mode in is clojure and ruby (all of python, octave, haskell, lisp, and  scheme include one).

tpapp:
> Should the existing inferior-julia be removed then?

Yes, I think so. I will open an issue.

> not having a full-featured inferior-julia builtin to julia-mode makes us relatively unique among the REPL-oriented languages

Possibly, but that's how it is. All solutions I am aware of are either WIP, have some serious limitations, or both. The way things are set up now, people can keep experimenting with various approaches to interacting with the REPL, while at the same time relying on this package to handle dealing with code.

(Incidentally, you list 8 languages including Julia, with 3 of them not having a REPL built-in to the default mode. Calling this *unique* is a somewhat heterodox use of the word.)

> if I could keep the inferior-julia prefix in order to ease possible future integration back into julia-mode

I am fine with anyone using `inferior-julia`, but please reflect on how committed you are to maintaining this. It would not be ideal to use up the name for something that ends up being abandonned.

Initially, maybe it would be preferable to use another name while you are working on this project, and in the meantime we can deprecate and remove inferior-julia.

non-Jedi:
>(Incidentally, you list 8 languages including Julia, with 3 of them not having a REPL built-in to the default mode. Calling this unique is a somewhat heterodox use of the word.)

Ha. I completely forgot about `ruby-mode` until I was about to hit submit and then didn't feel like doing too much editing. :man_shrugging: 

>I am fine with anyone using inferior-julia, but please reflect on how committed you are to maintaining this. It would not be ideal to use up the name for something that ends up being abandonned.

Good point. Fwiw, I am a bit more committed to this than when I submitted this PR--I think I can have something very nice to use with only 100-200 more lines of code. Since the primary reason I'm not using emacs-jupyter is the difficulty of hacking on it, this is very appealing to me.

PR #108, opened 2020-03-23
Add integration with project.el

non-Jedi:
This code is already used by eglot-jl to locate Julia projects and
will be useful in the future for other Julia packages (such as for
REPLs to decide which files they're associated with). Putting it in
the julia-mode package means it doesn't have to be redefined each time
it's used.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ install:
 
 script:
   - $EMACS --version
-  - $EMACS --batch -L . --eval "(and (>= emacs-major-version 24) (setq byte-compile-error-on-warn t))" -f batch-byte-compile julia-mode.el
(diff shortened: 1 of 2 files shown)


non-Jedi:
This doesn't byte-compile cleanly on Emacs 24 since `cl-defmethod` was added in 25.1, however all tests pass since that section is never executed since the `'project` feature was also added in 25.1. I'm sure there's a way to work around this, but I don't know how off the top of my head.

tpapp:
I am wondering if any package other than `eglot-jl` uses/needs something like this.

Also, I am reluctant to add a feature that only works on Emacs 25+, since we still support Emacs 24.

non-Jedi:
I found myself wanting this while working on #106, which is the reason I opened it.

berquist:
Can this be merged now that the minimum required Emacs version is 26.1?

non-Jedi:
@berquist What's your use-case for having this code in julia-mode?

tpapp:
Since I don't see a use case and the issue has been dormant for a while I am closing it. Feel free to comment or reopen if you disagree.

PR #110, opened 2020-03-24
Fix multiple type face highlighting. Add tests.

### LINKED ISSUE #107: Wrong highlighting with multiple `<:`'s
Hi!

I think julia-emacs is missing some font-locks when you have more than one `<:` in the same line. See for example this declaration:

Captura de Tela 2020-03-21 às 19 58 06

The first `Tuple` should also be highlighted.

tpapp:
Fixes #107.

### DIFF
--- a/julia-mode-tests.el
+++ b/julia-mode-tests.el
@@ -498,6 +498,14 @@ end")
     (julia--should-font-lock string 25 nil)
     (julia--should-font-lock string 26 nil)))
 
+(ert-deftest julia--test-typeparams-font-lock ()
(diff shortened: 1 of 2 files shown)


tpapp:
@giordano, @non-Jedi: this is a trivial PR but a review would be appreciated.

non-Jedi:
LGTM. :)

It occurs to me that since for `<:` to make sense semantically, both the lhs and rhs must be types. If we're font-locking the rhs, we should probably be font-locking the rhs with the same face. That's not a prerequisite for this PR though.

tpapp:
thanks!

PR #112, opened 2020-03-26
Remove highlighting for :: and <:.

### LINKED ISSUE #53: julia-mode and org html export with background
When html-exporting this Org mode example.org (with C-c C-e h o):

```
#+HTML_HEAD_EXTRA: <style type="text/css"> blockquote {background:#EEEEEE; padding: 3px 13px}    </style>
#+HTML_HEAD_EXTRA: <style type="text/css"> pre {background:#EEEEEE; padding: 3px 13px}    </style>

Some text...

#+BEGIN_EXAMPLE
:: or <: ?
#+END_EXAMPLE

#+BEGIN_SRC julia 
abstract type A end
struct S <: A
   x::T
end
#+END_SRC
```

I get this:
julia-mode-problem

I think that there is a problem in the way the "::" and "<:" char sequences are processed, in julia-mode.el one can read:
```
(defconst julia-font-lock-keywords
  (list
   ;; Ensure :: and <: aren't highlighted, so we don't confuse ::Foo with :foo.
   ;; (in Emacs, keywords don't overlap).
   (cons (rx (or "::" "<:")) ''default)
   ...
```

If I comment this line:
```
   ;; (cons (rx (or "::" "<:")) ''default) 
```
and export again example.org with Org, I get now:
julia-mode-ok
which is fine.

Commenting the line is quite brutal and certainly has side effect. 
Sure you will find a more elegant solution :-)

tpapp:
Removing this appears to be innocuous (ie nothing is changed). Fixes #53.

### DIFF
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -275,9 +275,6 @@
   ;; font-lock-builtin-face intentionally unused since any name from
   ;; names(Base) can be aliased in a baremodule.
   (list
-   ;; Ensure :: and <: aren't highlighted, so we don't confuse ::Foo with :foo.
-   ;; (in Emacs, keywords don't overlap).
-   (cons (rx (or "::" "<:")) ''default)
    ;; Highlight quoted symbols before keywords, so :function is not
    ;; highlighted as a keyword.
    (list julia-quoted-symbol-regex 1 ''julia-quoted-symbol-face)


tpapp:
I am not sure why this was there in the first place. 

@non-Jedi, @giordano, do you think that removing this is OK? Or should we create a face that does nothing?

PR #113, opened 2020-03-29
Give \ punctuation syntax outside of strings

### LINKED ISSUE #58: \( upsets paren matching

### LINKED ISSUE #59: (\)(2, 4) upsets auto-indent
The following
```
(\)(2, 4)
```
breaks auto-indent on subsequent lines.

The following, does *not*
```
\(2, 4)
```

I got *julia-mode.el* from elpa. But, the file is identical to the master branch here, except for a comment.

non-Jedi:
Fixes #58
Fixes #59

This commit also fixes a broken test for backslash indentation by
properly escaping backslashes in elisp strings and cleans up the code
for syntax-propertizing triple-quoted strings as a single string
rather than 3 strings.

### DIFF
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -85,6 +85,9 @@
     (modify-syntax-entry ?'  "." table)
     (modify-syntax-entry ?\" "\"" table)
     (modify-syntax-entry ?` "\"" table)
+    ;; Backslash has escape syntax for use in strings but
+    ;; julia-syntax-propertize-function sets punctuation syntax on it
+    ;; outside strings.
     (modify-syntax-entry ?\\ "\\" table)
 
     (modify-syntax-entry ?. "." table)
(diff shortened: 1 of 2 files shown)


julia-mode.el
```diff
@@ -313,33 +301,21 @@
...
-
 (defconst julia-syntax-propertize-function
   (syntax-propertize-rules
-   ("\"\"\""
-    (0 (ignore (julia-stringify-triple-quote))))
+   ;; triple-quoted strings are a single string rather than 3
+   ((rx (group ?\") ?\" (group ?\"))
+    (1 (let ((ppss (save-excursion (syntax-ppss (match-beginning 0)))))
+         (unless (or (nth 3 ppss) (nth 4 ppss))
+           (string-to-syntax "|"))))
```
tpapp: Is `string-to-syntax` needed? I thought that `syntax-propertize-rules` did that automatically for strings.

Also, can you please add some comments to each branch? Eg here we apply the syntax if we are inside a comment or a string.

Why do we need the second `group`, is it used for anything?

(I am not trying to nitpick here, just trying to make the code easier to maintain for people less familiar with Emacs Lisp and conventions.)

julia-mode.el
```diff
@@ -313,33 +301,21 @@
...
+    (2 (let ((ppss (save-excursion (syntax-ppss (match-beginning 0)))))
+         (unless (nth 4 ppss)
+           (when (nth 3 ppss)
+             (string-to-syntax "|"))))))
```
tpapp: same comments as above

tpapp:
@giordano, a review would be appreciated if you have the time.

julia-mode.el
non-Jedi: `string-to-syntax` is needed unless the `SYNTAX` part of `HIGHLIGHT1` is a string literal.

I'm pushing a commit with some additional comments.

The second group is used by the `(2 (...` expression to properly end triple-quoted-strings.

PR #114, opened 2020-04-01
Recognize whole triple-quoted string with regex instead of partial

### LINKED ISSUE #15: Highlight of triple quote string with escaped quote

non-Jedi:
Fixes #15

Feels kind of silly to make such a small change after just working on
this, but #113 made this fix much more obvious.

### DIFF
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -304,14 +304,17 @@
 (defconst julia-syntax-propertize-function
   (syntax-propertize-rules
    ;; triple-quoted strings are a single string rather than 3
-   ((rx (group ?\") ?\" (group ?\"))
+   ((rx (group ?\") "\"\""
+        (*? (or (not (any ?\\)) (and ?\\ anything)))
+        "\"\"" (group ?\"))
     ;; First " starts a string if not already inside a string (or comment)
-    (1 (let ((ppss (save-excursion (syntax-ppss (match-beginning 0)))))
-         (unless (or (nth 3 ppss) (nth 4 ppss))
-           (string-to-syntax "|"))))
+    (1 (unless (julia-syntax-comment-or-string-p
+                (save-excursion (syntax-ppss (match-beginning 0))))
+         (string-to-syntax "|")))
     ;; Last " ends a string if already inside a string
-    (2 (and (nth 3 (save-excursion (syntax-ppss (match-beginning 0))))
-            (string-to-syntax "|"))))
+    (2 (unless (julia-syntax-comment-or-string-p
+                (save-excursion (syntax-ppss (match-beginning 0))))
+         (string-to-syntax "|"))))
    ;; backslash acts as an operator if it's not inside a string
    ("\\\\"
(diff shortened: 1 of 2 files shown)


julia-mode.el
```diff
@@ -48,19 +48,15 @@
 
 (defcustom julia-indent-offset 4
   "Number of spaces per indentation level."
-  :type 'integer
-  :group 'julia)
```
tpapp, line 52: Can you please explain why the `:group` was removed?

julia-mode.el
```diff
@@ -48,19 +48,15 @@
...
+  :type 'integer)
 
 (defface julia-macro-face
   '((t :inherit font-lock-preprocessor-face))
-  "Face for Julia macro invocations."
-  :group 'julia-mode)
```
tpapp, line 57: also here, and the one below

tpapp:
Thank you very much for this. I actually prefer changes in small pieces so it is no problem.

@giordano, can you please review?

julia-mode.el
non-Jedi: Sure (there's also justification in the commit message). Since the group was declared with `:prefix "julia-"`, any `defcustom` that starts with "julia-" will automatically be in the group. So the `:group` is merely redundant.

tpapp: Thanks, I missed that. I didn't know about this feature of `defcustom`. Is this documented?

non-Jedi: Doesn't look like it. The manual is the only place I see `:prefix` mentioned, and it doesn't talk about implicitly including variables having the prefix.

https://www.gnu.org/software/emacs/manual/html_node/elisp/Group-Definitions.html

I'm aware of the feature because eglot makes use of it. Quickly checked few of emac's builtin progmodes and perl makes use of it, too. If it's an implementation detail, it's one that's at least relatively widely relied upon.

non-Jedi: Actually, it has nothing to do with `:prefix`. Bad assumption on my part. The elisp manual says:

>If a defcustom does not specify any :group, the last group defined with defgroup in the same file will be used. This way, most defcustom do not need an explicit :group.

https://www.gnu.org/software/emacs/manual/html_node/elisp/Variable-Definitions.html#Variable-Definitions

julia-mode.el
non-Jedi: This one was actually deleted because `julia-mode` was a group only created anonymously by these defgroup invocations. I'm assuming it was a mistake and they were meant to be members of `julia` group instead.

non-Jedi:
Don't merge this. It has some strange transient behavior I don't understand. When font-locking a buffer from scratch, it works exactly as expected, but when I insert a triple-quoted string into a buffer, everything after the triple-quoted string gets font-locked as a string for some reason until I e.g. `revert-buffer`.

non-Jedi:
Closing in favor of #133

PR #115, opened 2020-04-01
Indent lines after hanging operator even if previous line contains #

### LINKED ISSUE #19: indentation with the comment characters and the pipe operator
This indents just fine:
df = @linq df |>
transform(x = ismatch(r"[ ]*?([\d]{1,5})", :text)) |>
transform(y = :x \* 2)

This does not:
df = @linq df |>
transform(x = ismatch(r"#[ ]*?([\d]{1,5})", :text)) |>
transform (y = :x \* 2)

non-Jedi:
Because the previous implementation used a regex to detect whether the
hanging operator was in a comment or not, it did not detect hanging
operators following a # in a string.

Fixes #19

This also fixes an unrelated issue where the following didn't get properly indented. Haven't dug in to see what the underlying problem was. That's why the unrelated test had to be changed; the regex no longer applied since the buffer was getting properly indented.

```julia
f(x) =
x*
x
```


non-Jedi:
Please only review the last commit. This PR is based on #114 since it needs 8b37429 to byte-compile cleanly.

non-Jedi:
This is now ready for review independently.

non-Jedi:
@giordano or @FelipeLema could you please review this when you get a chance?

julia-mode.el
```diff
@@ -48,19 +48,15 @@
 
 (defcustom julia-indent-offset 4
   "Number of spaces per indentation level."
-  :type 'integer
-  :group 'julia)
+  :type 'integer)
```
FelipeLema, line 51: is there a reason why we're removing this custom-var from the `julia` group? maybe I missed a discussion about it

julia-mode.el
```diff
@@ -301,6 +297,21 @@
 (defconst julia-block-end-keywords
   (list "end" "else" "elseif" "catch" "finally"))
 
+(defsubst julia-syntax-comment-or-string-p (&optional syntax-ppss)
```
FelipeLema, line 300: are we moving this function from lower below to up here to group similar functions together?

on the matter, this looks like an internal function. maybe next time we'll rename these kind of functions as `julia--…`

julia-mode.el
FelipeLema: oh, I see... we're adding `:group 'julia` directly in the mode below. nvm

julia-mode.el
non-Jedi: `defsubst` invocations need to be before their invocation for the elisp to byte-compile cleanly. I think I moved the other two to be close to this one while I was at it.

julia-mode.el
non-Jedi: To clarify exactly what's going on here, custom variables are automatically added to the most recent `defgroup`.

tpapp: This is a nice feature, but since not everyone may be aware of it, can you please add a comment explaining this after the `defgroup`?

julia-mode.el
tpapp: It would be great if you could open an issue listing these functions, so that we don't forget.

julia-mode.el
```diff
@@ -388,7 +389,7 @@ symbol, gives up when this is not true."
       (while (and (not done) (< (point-min) (point)))
         (julia-safe-backward-sexp)
         (cond
-         ((looking-at (rx (or "import" "export" "using")))
+         ((looking-at (regexp-opt (list "import" "export" "using")))
```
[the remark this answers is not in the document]
tpapp: If I remember correctly `regexp-opt` is more efficient if it is applicable.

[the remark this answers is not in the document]
non-Jedi: That was my justification, but actually trying the functions, they generate the same regex strings in this case. I can revert if desired.

julia-mode.el
non-Jedi: Added comment in b15f13c

julia-mode.el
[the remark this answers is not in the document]
tpapp: No need, it is perfect as is.

PR #116, opened 2020-04-10
Correct the conditional for detecting if auctex is loaded

mattblackwell:
The `require` statement caused `auctex` to be loaded and then break `prettify-symbols-mode`. The `featurep` appears to be the correct way to check. Seems to pass the tests when I ran them.

### DIFF
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -774,7 +774,7 @@ strings."
 ;; (add-hook 'julia-mode-hook 'julia-math-mode)
 ;; (add-hook 'inferior-julia-mode-hook 'julia-math-mode)
 
-(when (require 'latex nil t)
+(when (featurep 'latex)
   (declare-function LaTeX-math-abbrev-prefix "latex")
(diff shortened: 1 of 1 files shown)


julia-mode.el
```diff
@@ -774,7 +774,7 @@ strings."
 ;; (add-hook 'julia-mode-hook 'julia-math-mode)
 ;; (add-hook 'inferior-julia-mode-hook 'julia-math-mode)
 
-(when (require 'latex nil t)
+(when (featurep 'latex)
```
giordano, line 777: Better to check
```suggestion
(when (featurep 'tex)
```

non-Jedi:
@rfourquet this has been reviewed and can be merged (pinging per zulip discussion).

rfourquet:
I confirm that this fixes the reported bug. I don't know why this was causing a bug but `featurep` seems to be the correct command here anyway. Thanks!

PR #123, opened 2020-05-02
Do not consider `:end` as block ending

ronisbr:
The word `:end` was being considered as a block ending. Hence, the
following code:

```julia
if a == :end
    r = 1
end
```

was being incorrectly indented:

```
if a == :end
r = 1
end
```

### DIFF
--- a/julia-mode-tests.el
+++ b/julia-mode-tests.el
@@ -441,7 +441,16 @@ r = 1
 end"
    "if a == :end
     r = 1
-end"))
+end")
+
+  (julia--should-indent
+   "if a == a[end-4:end]
+r = 1
+end"
+   "if a == a[end-4:end]
+    r = 1
+end")
+  )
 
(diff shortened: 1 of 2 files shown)


non-Jedi:
I think this is a reasonable workaround in the absence of proper
`backward-sexp` support. Note that this would be fixed by #6 which
would provide proper structural navigation (`julia-safe-backward-sexp`
would skip over).

ronisbr:
Thanks @non-Jedi for reviewing this!

I think I did what you asked. However, I am not very good with elisp, did I get everything right?

With respect to the test, I added a test considering `a[end-4:end]` which seems to me that can lead to more problemas than `a[1:end]` since we have a `end`.

tpapp:
@ronisbr: thanks for the PR.  I would like to ask @non-Jedi to kindly review again, then I am happy to merge.

PR #124, opened 2020-05-05
imenu fixes

li1:
Hi, I'm using this in my config, and while it's probably not perfect, I think it's better than the status quo:

This PR
- adds one-line fn support to imenu, including `@xyz` annotations: `@inline x(y) = ...`
- removes type (deprecated for newer Julia versions?)
- adds support for (mutable) struct instead: `mutable struct S{T} where T`
- requires `const =` and `^` s.t. const isn't accidentally parsed in strings etc.: `   const xzy =`

You can try it out by running `M-x imenu` in a `julia-mode` buffer.

### DIFF
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -700,15 +700,69 @@ Return nil if point is not in a function, otherwise point."
       (end-of-line)
       (point))))
 
+(defconst julia-imenu-private-multiline-fn
+  (rx-to-string
+   '(seq bol
+         (zero-or-more space)
+         (zero-or-one "@" (one-or-more alnum) space)
+         "function"
+         (one-or-more space)
+         (group "_" (minimal-match (zero-or-more anything)))
+         (or (seq (one-or-more space) "end") eol))))
+
+(defconst julia-imenu-public-multiline-fn
+  (rx-to-string
+   '(seq bol
+         (zero-or-more space)
+         (zero-or-one "@" (one-or-more alnum) space)
+         "function"
+         (one-or-more space)
+         (group (not (any "_")) (zero-or-more not-newline)))))
+
+(defconst julia-imenu-private-singleline-fn
+  (rx-to-string
+   '(seq bol
+         (zero-or-more space)
+         (zero-or-one "@" (one-or-more alnum) space)
+         (group "_"
+                (one-or-more (any "!" "." "_" alnum))
+                "(" (minimal-match (zero-or-more not-newline)) ")")
+         (minimal-match (zero-or-more not-newline)) space "=" space)))
+
+(defconst julia-imenu-public-singleline-fn
+  (rx-to-string
+   '(seq line-start
+         (zero-or-more space)
+         (zero-or-one "@" (one-or-more alnum) space)
+         (group (any "!" "." alnum)
+                (zero-or-more (any "!" "." "_" alnum))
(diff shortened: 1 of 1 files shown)


julia-mode.el
```diff
@@ -705,10 +705,11 @@ Return nil if point is not in a function, otherwise point."
...
   '(("Function (_)" "[ \t]*function[ \t]+\\(_[^ \t\n]*\\)" 1)
     ("Function" "^[ \t]*function[ \t]+\\([^_][^\t\n]*\\)" 1)
-    ("Const" "[ \t]*const \\([^ \t\n]*\\)" 1)
-    ("Type"  "^[ \t]*[a-zA-Z0-9_]*type[a-zA-Z0-9_]* \\([^ \t\n]*\\)" 1)
-    ("Require"      " *\\(\\brequire\\)(\\([^ \t\n)]*\\)" 2)
-    ("Include"      " *\\(\\binclude\\)(\\([^ \t\n)]*\\)" 2)
+    ("Function" "^[ \t]*\\(\\(@.+ \\)?[_a-zA-Z0-9]+\(.*\)\\) *= " 1)
+    ("Const" "^[ \t]*const \\([^ \t\n]*\\) *= " 1)
+    ("Struct" "^[ \t]*\\(\\(mutable \\)*struct [^ \t\n]*\\)" 1)
+    ("Require" " *\\(\\brequire\\)(\\([^ \t\n)]*\\)" 2)
```
non-Jedi, line 712: It's hard to tell exactly which lines/regexes were changed in this commit and which are just whitespace cleanups. Could you please separate out the pure whitespace cleanups into a separate commit? Or are they all changing, and my eyes are just failing me?

li1: Haha, I can see (or not?) where you're coming from.

All lines except `Require` & `Include` changed. `Require` and `Include` only contain a whitespace change.

I can try to add a short comment for each line on how the regex works, because I'm not sure looking at the individual commits is much help?

tpapp: It would help greatly if you could rewrite the changed regexps using `rx`. See https://github.com/JuliaEditorSupport/julia-emacs#contributing

li1: Rewrote to `rx` — I think this also makes the diff much clearer.

tpapp: Thanks for the rewrite, indeed they are much cleaner. I wonder if it would make sense to add some tests (basically just testing that the regexps match what they need to). I realize that the package currently does not have tests for these, but it would make it easier to maintain.

li1: I agree, however probably won't have time anytime soon to add them 🙁

ronisbr:
Hi! Just to mention that I used this patch and it improved a lot! A lot of my functions had some kind of annotations and now I can successfully use `imenu` to jump to them.

tpapp:
Thanks for the feedback. I am looking forward to merging this, but I would really appreciate having some tests. The PR fixes functionality that was broken before, and adds new features. At some point these may break again; having tests would make these things much easier to maintain.

I am not trying to be difficult, just considerate of our great-grandchildren using Emacs 102.0 with Julia 7.0 in 2097. They will thank us.

tpapp:
Again, it would be great to merge this, we just need tests. CI is running again.

Issue #132, opened 2020-07-09
Highlight the contents of cmd macros as if they were standard code


MasonProtter:
Continued from https://github.com/JuliaLang/julia/issues/36590.

Currently, string and cmd macros are often used for syntax extensions. While the contents of a string macro might sometimes be code or sometimes a special form of string like LaTeXStrings.jl, cmd macros on the other hand seem to always contain code. However, we highlight their contents as if they were strings. 

I think it'd be nice if we just applied regular julia syntax highlighting to the contents of a cmd macro, or even provided hooks so that people can customize the highlighting of specific string / cmd macros.

non-Jedi:
I wonder if this functionality could be a minor-mode on top of julia-mode itself. It seems a bit icky to special-case different string-macro names to have/not-have this behavior based on common usage rather than language semantics. You might not even have to modify julia-mode to create a "hook" to do this...

I don't think I've ever seen a cmd macro in the wild. Do you have some examples of them?

MasonProtter:
> I wonder if this functionality could be a minor-mode on top of julia-mode itself. It seems a bit icky to special-case different string-macro names to have/not-have this behavior based on common usage rather than language semantics. You might not even have to modify julia-mode to create a "hook" to do this...

I guess my thought was to just treat the contents of a cmd block as if it were a completely separate block of julia code, both for highlighting and possibly even indentation. Having custom overrides to use e.g. python syntax highlighting in a pycall string macro would be a bonus, but I don't think necessary.

> I don't think I've ever seen a cmd macro in the wild. Do you have some examples of them?

They're not actually used a ton, but on an ecosystem level, I think it might make sense to encourage their use over string macros for containing code, rather than just funky strings. One example would be https://github.com/JuliaInterop/MathLink.jl

Amusingly enough, another example is the undecorated cmd syntax: https://github.com/JuliaLang/julia/blob/24f033c9517cd186448acc3bededa1a64eaad09f/base/cmd.jl#L388

tpapp:
I am sorry, but I still don't get what a cmd macro is (ie how it is different from a string macro). The Julia issue linked looks like a feature request which may not even make it, so it is unclear to me what is requested here. An example would help.

MasonProtter:
> I am sorry, but I still don't get what a cmd macro is (ie how it is different from a string macro).

[...]

PR #133, opened 2020-07-09
Apply proper syntax to triple-quoted cmds

### LINKED ISSUE #15: Highlight of triple quote string with escaped quote

### LINKED ISSUE #119: Wrong highlighting when multi-line string is closed just before an escaped quote
Hi!

The highlighting engine does not understand that a multi-line comment section (`"""`) was ended if it ends just after an escaped quote:

Captura de Tela 2020-04-21 às 17 42 23

If I add a space between `\"` and `"""`, then everything works.

non-Jedi:
This is necessary cleanup before #132 could be attempted and should also make triple-backticks work correctly which has never been true before.

This also appears to have fixed all the corner-cases with escapes in strings.

Closes #15, closes #119

### DIFF
--- a/julia-mode-tests.el
+++ b/julia-mode-tests.el
@@ -532,8 +532,17 @@ end")
     (julia--should-font-lock string 74 font-lock-type-face) ; B
     ))
 
+(ert-deftest julia--test-single-quote-string-font-lock ()
+  "Test that single quoted strings are font-locked correctly even with escapes."
+  ;; Issue #15
+  (let ((s1 "\"a\\\"b\"c"))
+    (julia--should-font-lock s1 2 font-lock-string-face)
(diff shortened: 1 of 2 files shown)


tpapp:
@giordano @FelipeLema @rfourquet if any of you has time, please review this, I would like to merge.

tpapp:
Yes, I just resolved that manually (I am surprised that `git` hung up on it, there was no actual conflict), and will merge as soon as tests finish.

@non-Jedi: as always, thanks for the excellent work!

PR #134, opened 2020-07-13
font-lock ccall as builtin

### LINKED ISSUE #125: font-lock-builtin-face for true builtins
In #102 I removed usage of `font-lock-builtin-face` for a wide range of types; I'm now wondering if it's a mistake to do so. I had assumed that those types wouldn't be available in a `baremodule`, forgetting that `Core` is still imported for a baremodule. They can still be overridden though.

<del>At the very least, `true` and `false` should be font-lock-builtin-face since they cannot be aliased or overridden and are available in `baremodule`.</del> `ccall`, too I suppose.

So proposed `font-lock-builtin-face` candidates are `ccall`, <del>`true` and `false`</del> plus `names(Core)`:

- `<:`
- `===`
- `AbstractArray`
- `AbstractChar`
- `AbstractFloat`
- `AbstractString`
- `Any`
- `ArgumentError`
- `Array`
- `AssertionError`
- `Bool`
- `BoundsError`
- `Char`
- `Core`
- `Cvoid`
- `DataType`
- `DenseArray`
- `DivideError`
- `DomainError`
- `ErrorException`
- `Exception`
- `Expr`
- `Float16`
- `Float32`
- `Float64`
- `Function`
- `GlobalRef`
- `IO`
- `InexactError`
- `InitError`
- `Int`
- `Int128`
- `Int16`
- `Int32`
- `Int64`
- `Int8`
- `Integer`
- `InterruptException`
- `LineNumberNode`
- `LoadError`
- `Main`
- `Method`
- `MethodError`
- `Module`
- `NTuple`
- `NamedTuple`
- `Nothing`
- `Number`
- `OutOfMemoryError`
- `OverflowError`
- `Ptr`
- `QuoteNode`
- `ReadOnlyMemoryError`
- `Real`
- `Ref`
- `SegmentationFault`
- `Signed`
- `StackOverflowError`
- `String`
- `Symbol`
- `Task`
- `Tuple`
- `Type`
- `TypeError`
- `TypeVar`
- `UInt`
- `UInt128`
- `UInt16`
- `UInt32`
- `UInt64`
- `UInt8`
- `UndefInitializer`
- `UndefKeywordError`
- `UndefRefError`
- `UndefVarError`
- `Union`
- `UnionAll`
- `Unsigned`
- `Vararg`
- `VecElement`
- `WeakRef`
- `applicable`
- `eval`
- `fieldtype`
- `getfield`
- `ifelse`
- `invoke`
- `isa`
- `isdefined`
- `nfields`
- `nothing`
- `setfield!`
- `throw`
- `tuple`
- `typeassert`
- `typeof`
- `undef`

Most of these aren't privileged in that they can't be aliased, but they are privileged in that they are availabe in every module by default, and they are very unlikely to be overridden. I'd lean slightly towards font-locking them specially.

EDIT: after sleeping on it, I think I've come back around to thinking that only `ccall` should be `font-lock-builtin-face`. `true` and `false` should probably stay as `font-lock-constant-face`.

non-Jedi:
Closes #125 

Since the semantics of `ccall` are so similar to a function call, it makes sense to highlight it as a builtin rather than as a keyword.

### DIFF
--- a/julia-mode-tests.el
+++ b/julia-mode-tests.el
@@ -532,6 +532,12 @@ end")
     (julia--should-font-lock string 74 font-lock-type-face) ; B
     ))
 
+(ert-deftest julia--test-ccall-font-lock ()
+  (let ((s1 "t = ccall(:clock, Int32, ())"))
+    (julia--should-font-lock s1 5 font-lock-builtin-face)
+    (julia--should-font-lock s1 4 nil)
+    (julia--should-font-lock s1 10 nil)))
+
 ;;; Movement
 (ert-deftest julia--test-beginning-of-defun-assn-1 ()
   "Point moves to beginning of single-line assignment function."
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -273,7 +273,8 @@
    (cons julia-macro-regex ''julia-macro-face)
    (cons
     (regexp-opt
-     '("true" "false" "C_NULL" "Inf" "NaN" "Inf32" "NaN32" "nothing" "undef" "missing")
+     ;; constants defined in Core plus true/false
+     '("true" "false" "Cvoid" "Inf" "NaN" "Inf32" "NaN32" "nothing" "undef" "missing")
      'symbols)
     'font-lock-constant-face)
    (cons "ccall" 'font-lock-builtin-face)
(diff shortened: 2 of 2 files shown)


non-Jedi:
I've also cherry-picked the stylistic fixes from #114 into this branch.

Issue #142, opened 2020-08-30
Wrong highlighting and indentation after a char that is an escaped quote


ronisbr:
Hi!

If we have a char that is a quote `'`, then the highlight and indentation is not correct:

<img width="77" alt="Captura de Tela 2020-08-29 às 22 13 09" src="https://user-images.githubusercontent.com/1068295/91648981-d32f7580-ea44-11ea-95d0-d3debd39167d.png">

recri:
Same thing, after this line:
```
            elseif c in ('\'', '\"', '\`')
```
entire rest of source file is rendered as part of a string.

recri:
Opened a pull request that appears to fix this.  `julia-char-regex` looked very obscure, but it was only wrong.  Syntax table is unchanged.

PR #143, opened 2020-09-08
Rec/quoted quote

recri:
I think this solves https://github.com/JuliaEditorSupport/julia-emacs/issues/142 by simply rewriting `julia-char-regex`.  The regex didn't actually recognize `'\'`' as a character constant as it was written.  It could still be wrong.   Added a test for my particular case.

Sorry, I should have rebased or something to squash  the intermediate commits.

### DIFF
--- a/julia-mode-tests.el
+++ b/julia-mode-tests.el
@@ -88,13 +88,6 @@ matching line or end of match if END is non-nil.  Optional ARG is passed to FUN.
                                       (point-at-bol)))
                            ,to)))))
 
-(defmacro julia--should-regex-char-const (text)
-  "(string-match julia-char-regex `text') should return text"
-  `(with-temp-buffer
-     (julia-mode)
-     (and (= 0 (string-match julia-char-regex ,text))
-	  (string= ,text (match-string 0 ,text)))))
-
 ;;; indent tests
 
 (ert-deftest julia--test-indent-if ()
@@ -489,6 +482,18 @@ identity"
 a = \"#\" # |>
 identity"))
 
+(ert-deftest julia--test-indent-quoted-single-quote ()
+  "We should indent after seeing a character constant containing a single quote character."
+  (julia--should-indent "
+if c in ('\'')
+s = \"$c$c\"*string[startpos:pos]
+end
+" "
+if c in ('\'')
+    s = \"$c$c\"*string[startpos:pos]
+end
+"))
+
 ;;; font-lock tests
 
 (ert-deftest julia--test-symbol-font-locking-at-bol ()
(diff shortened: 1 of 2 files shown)


julia-mode.el
```diff
@@ -110,8 +110,9 @@
             (syntax whitespace)
             bol)
         (group "'")
-        (or (repeat 0 8 (not (any "'"))) (not (any "\\"))
-            "\\\\")
+        (or (not (any "\\'"))
+	    (seq "\\" (or (repeat 1 9 (not (any "\\'")))
+			  (any "\\'"))))
```
non-Jedi, line 115: Does this match [e.g. `'\n'`, `\0`, `\x7f`, `\177`](https://docs.julialang.org/en/v1/manual/strings/#man-characters)? It doesn't seem like it would. Probably worth adding a test for those.

julia-mode-tests.el
```diff
@@ -482,6 +482,18 @@ identity"
 a = \"#\" # |>
 identity"))
 
+(ert-deftest julia--test-indent-quoted-single-quote ()
```
non-Jedi, line 485: Please add `julia--should-font-lock` tests for examples from #142.

non-Jedi (changes requested):
If the tests I suggested are added and pass, this LGTM.

julia-mode.el
recri: They all fit into the "(1 to 9 (not (backslash or singlequote))) after backslash" clause.  I could be more prescriptive, if you like.

julia-mode-tests.el
```diff
@@ -620,6 +632,10 @@ identity"))
     (julia--should-font-lock s1 4 nil)
     (julia--should-font-lock s1 10 nil)))
 
+(ert-deftest julia--test-char-const-font-lock ()
+  (dolist (c '("'\\''" "'\\\"'" "'\\\\'" "'\\010'" "'\\xfe'" "'\\uabcd'" "'\\Uabcdef01'" "'\\n'" "'\\alpha'" "'a'"  "'z'"))
+    (julia--should-font-lock c 1 font-lock-string-face)))
```
non-Jedi, line 637: Need to test that the string-face ends at the end of the char literal.

julia-mode.el
non-Jedi: That makes sense. This is fine. If you wanted to, you could add a couple comments with more explanation of the different forms of char literals and how they're matched, but I don't think that should block merging this.

tpapp: Yes, explanations would be great if @recri has the time. Also, indentation of `(seq ...` looks a bit off.

julia-mode.el
```diff
@@ -109,11 +109,17 @@
...
+        (group "'")                     ; start single quote of character constant
+        (or                             ; two alternatives
+         (not (any "\\'"))              ; one character, not single quote or backslash
+	     (seq "\\"                      ; sequence of a backslash followed by ...
+              (or                       ; four alternatives
+               (any "\\'\"?abfnrtv")    ; single character escape
+               (repeat 1 3 (any "0-7")) ; octal escape
+               (seq 'x' (repeat 1 8 hex))     ; hex escape
+               (seq 'u' (repeat 1 4 hex))     ; unicode escape
+               (seq 'U' (repeat 1 8 hex)))))  ; extended unicode escaple
```
tpapp, line 121:
```suggestion
        (or                             ; two alternatives
         (not (any "\\'"))              ; one character, not single quote or backslash
         (seq "\\"                      ; sequence of a backslash followed by ...
              (or                       ; four alternatives
               (any "\\'\"?abfnrtv")    ; single character escape
               (repeat 1 3 (any "0-7")) ; octal escape
               (seq 'x' (repeat 1 8 hex))     ; hex escape
               (seq 'u' (repeat 1 4 hex))     ; unicode escape
               (seq 'U' (repeat 1 8 hex)))))  ; extended unicode escaple
```

recri:
Sorry, I seem to have crossed the git gods somehow.

tpapp:
Thanks for all the work! LGTM, I will merge when @non-Jedi approves.

julia-mode.el
```diff
@@ -109,10 +109,17 @@
             (syntax open-parenthesis)
             (syntax whitespace)
             bol)
-        (group "'")
-        (or (repeat 0 8 (not (any "'"))) (not (any "\\"))
-            "\\\\")
-        (group "'"))))
+        (group "'")        ; start single quote of character constant
+        (or                ; two alternatives
+         (not (any "\\'")) ; one character, not single quote or backslash
```
non-Jedi, line 114: `'''` is a valid Char constant for `'\''`. This doesn't match that does it?

recri: Amazing, that must have been the only quoted single quote that the original code recognized, and the only one that's not documented in the julia manual.

This last pull also tightens up the \x syntax, julia only permits two hex digits, C permits as many as you write though not guaranteeing what will happen to the excess..

non-Jedi (approved):
Thanks! LGTM.

Issue #111, opened 2020-03-24
Indentation of map do statement


hcarlsso:
When writing a map-do statement, indentation is correct:
```
map(1:3) do x
    x
end
```
However, if writing this in a function call yields
```
f(
    map(1:3) do x
    x
    end
)
```
where the statements inside the do-block is on the same level as `map`. Shouldn't it be like the following?
```
f(
    map(1:3) do x
        x
    end
)
```

tpapp:
I think this is a dup of #11, but will keep it open to see if fixing that closes it.

PR #152, opened 2021-01-19
Allow block indentation inside of parentheticals

hexane360:
I've tweaked the indentation code slightly, so that blocks are indented correctly inside of parentheses and brackets. I've also added a couple of test cases.

Here's some examples:
```julia
block_in_func(begin
                  arg = 2,
                  begin
                      test
                  end
              end)

var = func(arg1, arg2, begin
               arg3
           end)

begin
    var = [aligned,  # works with brackets and parens
           arg,
           begin
               statement
               arg
           end,
           arg
           ]
    func(
        indented,
        1 +  # works with hanging operators
            begin
                arg
            end,
        arg
    )
end
```

This should fix a large portion (though not all) of the complaints in #11.

### DIFF
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -459,7 +459,20 @@ Do not move back beyond MIN."
     (and pos
 	 (progn
 	   (goto-char pos)
-	   (+ julia-indent-offset (current-indentation))))))
+	   (+ julia-indent-offset (julia-block-open-indentation))))))
+
+(defun julia-block-open-indentation ()
+  "Get the current indentation or the start of a parenthetical block."
+  (save-excursion
+    ;; narrow to one line to only search syntax on that line
+    (narrow-to-region (line-beginning-position) (line-end-position))
+    (prog1
+        (condition-case nil
+            (progn
+              (backward-up-list)
+              (1+ (current-column)))
+          (error (current-indentation)))
+      (widen))))
 
 (defcustom julia-max-block-lookback 20000
   "When indenting, don't look back more than this many characters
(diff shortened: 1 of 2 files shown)


hexane360:
BTW, here's the indentation given for some of the test cases in issue #11:

```julia
function1(a, b, c # correct, alignment style
          d, e, f)
function2( # correct, indent style
    a, b, c
    d, e, f)
for i in Float64[1, 2, 3, 4 #correct, alignment style
                 5, 6, 7, 8]
end
for i in Float64[ #correct, indent style
    1, 2, 3, 4
    5, 6, 7, 8]
end
a = function3(function () #correct, alignment style
                  return 1
              end)
a = function4( #correct, indent style
    function ()
        return 1
    end
)
```
```julia
longfunctionname(;  # I don't know if we should special-case this
                 hello=nothing,
                 world=nothing,
                 )
longfunctionname(
    hello=nothing,
    world=nothing,
)
```

```julia
hl = Highlighter((h,data,i,j)->begin
                     id = div(i-1, 3)
                     sim_id = id*num_cols + j
                     if sims_status[sim_id] == 1
                         return crayon"black bg:green"
                     else
                         return crayon"bg:yellow"
                     end
                 end)
```

```julia
f(a, b,
  c) do i
      sin(i)  # i agree this could still look better
end
```

And issue #111:
```julia
f(
    map(1:3) do x
        x
    end
)
```

ronisbr:
This is awesome!

@tpapp @non-Jedi can you please take a look at this PR? This modification is very important to julia-emacs. It fixes some annoying issues.

julia-mode.el
```diff
@@ -459,7 +459,20 @@ Do not move back beyond MIN."
...
+  (save-excursion
+    ;; narrow to one line to only search syntax on that line
+    (narrow-to-region (line-beginning-position) (line-end-position))
+    (prog1
+        (condition-case nil
+            (progn
+              (backward-up-list)
+              (1+ (current-column)))
+          (error (current-indentation)))
+      (widen))))
```
tpapp, line 475: instead of `(widen)`, perhaps wrap the narrowing in a `(save-restriction ...)` block?

tpapp (changes requested):
Very nice PR, just requested a minor style change.

tpapp:
@hexane360: thanks for this very nice PR, and apologies for the late review. I will wait for a review from @non-Jedi (if he has the time) and them I am happy to merge, I just suggested one minor change.

@ronisbr: thanks for the ping.

ronisbr:
@tpapp thanks for the quick response! :)

ronisbr:
@tpapp is it possible to accept this without that proposed modification and then fix if any problem occurs? Is there any big problem with `(widen)`?

hexane360:
I just went ahead and wrote up the style change. This should be ready to accept now.

tpapp:
@hexane360: thanks.

@non-Jedi, @giordano, @FelipeLema: if any of you have time for a review, it would be appreciated, then I would merge.

FelipeLema (approved):
looks good (Y)

giordano (approved):
Looks good to me!

Issue #97, opened 2019-12-21
volunteers needed for reviewing PRs


tpapp:
I would like to aim for faster turnaround times for PRs submitted to this repository. I am always happy to review, but since I am hardly an Emacs Lisp expert, I would appreciate if others could help out.

I am not comfortable with pinging people without making sure they are OK with it. **So if you have some Emacs Lisp experience and would be willing to help out, please respond here**.

 Please do this even if you are a maintainer already with commit access, it is unclear to me who is active from a rather long list.

giordano:
I'm happy to be pinged

tpapp:
And just to clarify: reviews are welcome from anyone, regardless of whether it was requested on Github. So if someone is willing to help out and has the time for a review, feel free to just do it.

non-Jedi:
I have a modicum of elisp experience. Feel free to ping me for reviews.

FelipeLema:
I maintain three packages @ melpa, all have been peer-reviewed so I got a good grip of maintainable elisp.

Feel free to ping me for reviews.

ajhalme:
I don't know much about Emacs Lisp, but I'm interested in the project.

I'd be happy to take a look at a PR now and then.

PR #159, opened 2021-08-21
Fix issue in keyword indentation with compound keywords

ajhalme:
While playing around with Julia macros, I noticed that `abstract type` definitions inside a begin-end block were breaking the indentation in my Emacs.

I've never written anything in Emacs Lisp before, but the broken indentation was annoying me enough that I figured out how to fix it.

On closer inspection, I determined that the issue is with `julia-last-open-block-pos` and `julia-at-keyword`. The block position function determines the nesting level by scanning code preceding `point` one s-expressions at a time, keeping count of matched block starting and ending keywords. The `julia-at-keyword` function tries to match on keywords listed in `julia-block-start-keywords`, but unfortunately the matching doesn't quite work with _compound keywords_ made up of more than one word.

Because `julia-at-keyword` uses `(current-word)`, and because `abstract` and `type` are parsed as separate words, the matching misses the compound keyword `abstract type` (and `primitive type` and `mutable struct`). This results in the begin/end counting algorithm missing the compound keyword block beginnings, breaking the indentation.

This PR fixes this issue by introducing a new matching function specifically for compound keywords, `julia-at-compound-keyword`. This function looks at the current word and the one preceding it, and determines whether `point` is on the second word of a compound keyword. To make use of this function, I propose splitting the keyword list in two parts, with the compound words in their own list. The new function is applied as an additional clause in the block begin/end counting function.

I've included two tests that demonstrate the desired behaviour.

The performance cost from running the extra function should be negligible, though it is called for every s-expression step inside the matching algorithm.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Fix issue in keyword indentation with compound keywords ([#159](https://github.com/JuliaEditorSupport/julia-emacs/pull/159))
+
 # 0.4
 
 - increase lookback ([#98](https://github.com/JuliaEditorSupport/julia-emacs/pull/98)), fixes [#5](https://github.com/JuliaEditorSupport/julia-emacs/issues/5)
--- a/julia-mode-tests.el
+++ b/julia-mode-tests.el
@@ -581,6 +581,34 @@ var = func(begin
            end
            )"))
 
+(ert-deftest julia--test-indent-block-compound-keywords ()
+  "We should indent on compound keywords inside a block."
+  (julia--should-indent "
+begin
+abstract type MyType end
+test
+end" "
+begin
+    abstract type MyType end
+    test
+end"))
+
+(ert-deftest julia--test-indent-block-compound-keywords-with-nesting ()
+  "We should indent on compound keywords and again inside them."
+  (julia--should-indent "
+begin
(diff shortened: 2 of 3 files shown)


tpapp (approved):
LGTM.

tpapp:
@giordano @non-Jedi @FelipeLema If you have time a review would be appreciated.

Currently CI is not running, I ran the tests locally and the two font-lock failures seem unrelated.

tpapp:
CI failure was local to a quirk my setup, sorry for the noise. Tests actually run clean.

tpapp:
Closing and reopening to trigger CI.

tpapp:
Incidentally, even if you have

> never written anything in Emacs Lisp before

you write nice and clean code, so if you would be willing to review PRs, just comment in #97.

FelipeLema:
ohmydays I was looking forward to this... do you need any help to complete this?

∧_∧ 
　／＼(`･ω･)／＼ 
(　●　⊂　　 ⊃●　) 
　＼／/ 　 < ＼／ 
　∧_∧/⌒＼)∧_∧ 
（ ･ω･∩　∩･ω･ ) 
⊂　　 ノ　ヽ　　 つ 
`/　　/　　 ｜　　< 
(_ノヽ)　　 (_ノ＼_)

PR #163, opened 2021-08-23
Add logo, add section on testing, README tweaks

ajhalme:
I noticed that [the *other* editor support project](https://github.com/JuliaEditorSupport/julia-vim) has a nice mashup logo on the landing page. I feel like a nice logo might be just the thing that this project needs.

My Julia + Emacs mashup logo proposal attached.

I also propose adding a section on tests in the README. Last weekend, when I was thinking about submitting a PR, discovering how easy it is to run the test suite was really encouraging. Highlighting the test suite might encourage others to engage as well.

While I was at it, I also did some minor tweaks to the README document overall.

I appreciate that a README PR was recently merged, but I feel like this PR still improves on the document enough to warrant consideration. That said, I'm not trying to push for anything in particular -- happy to make changes or even abandon. Not an urgent patch by any means.

julia-mode-logo

PR #164, opened 2021-08-25
Add svg logo

ajhalme:
This PR replaces the png logo with an svg, as a follow-up to #163.

Not sure what I fumbled with the file yesterday, but this one seems to have the right scale for using as is.


ajhalme:
I figured out enough Luxor to redo the logo.

I went with some of the parameters of the original, but due to some Bézier curve weirdness, the output is a little edgier. But I kind of like it that way. It's also _slightly_ larger than the previous one -- do let me know it it's too massive, and I can try and figure out some scaling. Happy to tweak everything else in the Julia code as well.

I've included the script in a new Julia project in `logo/`. I hope I set that up right. The script generates an svg in the subdirectory root. I also linked the README there directly, removing the old file.

Finally, I includes a separate LICENSE file in `logo/`, based on [the one in the main logo repo](https://github.com/JuliaLang/julia-logo-graphics/blob/master/LICENSE.md). I think declaring a CC (BY NC SA) license would be a reasonable compromise here.

tpapp:
Thanks for the work, but this is a log of baggage just to get a logo. Could we just generate a simple one from Luxor and call it a day?

ajhalme:
Agreed about excess baggage. I just wanted to try and do the right thing. One of those threads to avoid pulling on too hard.

I was able to reach Nicolas and he's fine with his design being reused here.

In any case, I've picked up some more Luxor skills, and I went ahead and redid the logo. I changed the gradient to use Julia colours and drew a new curve to serve as a reference to Emacs. I also parameterised the draw calls so the image is easier to scale, if necessary.

I've left most of the baggage behind, and instead added a few lines to the main LICENCE file.

Finally, I squashed all my commits into one.

How does that look?

https://github.com/ajhalme/julia-emacs/tree/logo-svg

ajhalme:
**tl;dr** 
Nicolas Petton, author of the Emacs logo referenced in the logo currently on master, kindly agreed to let me "reuse and adapt the logo under CC BY NC SA in this context as it has a reference to Emacs".

As such, there should be no licence issue with either the logo on master or the one on this branch. Both works can be included in this project and can be distributed under the terms of the CC "BY-NC-SA" 4.0 licence.

**
> Since the proposed logo is a derivative of both the Emacs logo and the Julia logo, it is covered by copyrights to both works. The GPL and CC-BY-NC are not compatible since the CC-BY-NC imposes "additional restrictions" on distribution of the work; thus distributing this derivative work violates the Emacs logo's license.

I think it would be more precise to say that the logo I originally put together had an unclear legal standing, because at the time the work would have had to be simultaneously licensed for use under two incompatible share-alike licences.

As the author of the pull request that introduced the logo, I implicitly claimed authorship/copyright of the work, even though strictly speaking I did not at the time have this authority based on the publicly available licences to the underlying works. Charmed by the immense artistic value of my mashup and the positive experience of a previous PR submission, I pressed on not realising that there was this licencing issue. This was the mistake on my part, I should have known better. I should not have contributed content to the project that I was not fully in charge of. I apologise for that.

On realising my error, I ventured to rectify to the situation by getting in touch with the author of the Emacs logo, Nicolas Petton. Nicolas kindly agreed to let me "reuse and adapt the logo under CC BY NC SA in this context as it has a reference to Emacs".

This clarifies the legal standing of the logo currently on master, as it can be seen as a mashup that respects the terms and conditions of the CC (BY NC SA) licence, under which the Julia logo and now Petton's Emacs logo can be used here. The intention is that the mashup itself respects the same licence -- another thing this PR tries to clarify.

To be clear, there is no GPL violation here. Contributors retain authorship of the contributions they make, so Petton should be well within his rights to re-license his logo. For the purpose of the derived works here, there is no longer a need to "go through" Emacs and the GPL to gain access to the asset.

> This logo (and the one currently committed to the repo) cannot be legally distributed and should be removed.

The way I see it is that the point of a licence is to specify the terms and conditions under which a work can be used. Specifically, as is useful in many cases involving software with many contributors, one doesn't need to get any additional permissions or indeed be in contact with the authors in any way. The purpose of a licence is to clarify and simplify authorised usage.

But sometimes existing licences are not enough. Sometimes new terms and conditions can be agree to, often using other existing licences as a convenience. This is why including the author and their contact details with the licence is so important. When there are licensing concerns, it is entirely possible that other terms and conditions can be agreed to. Of course, authors have no obligation to licence their work in any way.

"This logo cannot be legally distributed" is quite a strong, seemingly definite claim. This whole thing appeared much more fuzzy to me then. It would have been perfectly reasonable to exclude the logo from this project on the grounds of its unclear legal standing, but that doesn't mean that the logo cannot _ever_ be distributed _anywhere_ in _any_ context under _any_ terms.

Further, I'm not even sure if the new logo on this branch is any longer a derived work of Petton's Emacs logo. This one makes no use of Petton's image, uses a different method for the construction of the image, features different shapes, uses Julia colours, etc. There is a limit somewhere. Indeed, I went for this second version partly because I wasn't sure at the time if I would be able to reach Petton to agree on different terms for the use of his image.

> Doing this by the book would actually require rewriting the repo's git history to remove the copyright-violating logo, but we probably wouldn't get in trouble if we simply added a commit removing it.

Ah, yes, the book. But which book? :)

If I have committed a crime by falsely (implicitly) claiming copyright on a work when I strictly speaking wasn't in the clear to do so, I don't think even "undoing the damage" by removing the material from the git history would be enough. If I had highly valuable copyright covered content that somebody else shares even for a few days, I could probably pursue legal action against them even if they ceased to distribute. In other words, the milk has already been spilled.

I'm not a lawyer, but I think the only way to really ameliorate the situation is to establish terms of use that the actual copyright holder is fine with. This is exactly what I did. We didn't explicitly cover in our exchange the period of time between when I made the new work available and when he agreed to the new terms, but my understanding is that his blessing is retroactive in this way as well. 

In short, Petton's approval to use his logo should clear me of any wrongdoing and should keep you out of any hypothetical trouble as well.

> Alternatively, this problem would go away if [Stefan] was willing to release the Julia logo under a permissive license?

I agree, but that would be both silly and now unnecessary. GPL is not a great choice for art assets.

StefanKarpinski:
I guess the "problem" with the Julia logo license is that it restricts commercial usage? I'm afraid I'm not willing to relax that requirement. I'd be perfectly happy to grant any sort of exemption to this particular logo, but that's not the issue: the issue is that the Emacs logo's GPL license demands that the entire derived work, including the Julia logo, be licensed under terms that satisfy the GPL, including letting anyone use the Julia logo for commercial purposes without requiring permission. The real issue here is the restrictiveness of the Emacs logo, which I guess there's not much we can do about, unfortunately. I'd be willing to grant a GPL license for the use of the combined Julia-dots-Emacs logo as long as it retains the Emacs part and isn't incorporated in anything else, but I'm not sure these requirements are GPL-compatible and I think we'd need legal counsel to figure out if any of this is legally valid...

StefanKarpinski:
Here's an interesting thought: I could grant an AGPLv3 license for the use of the Julia logo, which would allow combining it with anything GPL-licensed, but would, by my reading, require anyone using the logo for commercial usage to provide all the source code implementing their software and/or service 😈

tpapp:
I think it would be best to revert that PR and remove the logo for now. Sorry if I was too hasty merging it.

Issue #148, opened 2020-11-25
Highlighting `in` in multi-iterable `for`


phipsgabler:
I either get this: 

![Screenshot_2020-11-25_18-50-34](https://user-images.githubusercontent.com/7878215/100264539-40bd4600-2f4f-11eb-81ac-bc747448ace7.png)

Or this:

![Screenshot_2020-11-25_18-50-56](https://user-images.githubusercontent.com/7878215/100264555-4581fa00-2f4f-11eb-89a7-c3b88dd05910.png)

non-Jedi:
This is a tough one that would require an actual parser to do right. But you could probably modify `julia-forloop-in-regex` to handle the simple cases; at least the single-line cases.

PR #169, opened 2022-03-11
Improve highlighting of `in`.

mcol:
This highlights both `in` keywords in multi-iterable `for` loops (#148), and adds highlighting in the cases where `in` is used to check if an element is contained in an array, as in `if 2 in [1, 2, 3]`.

### DIFF
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -189,8 +189,8 @@
 (defconst julia-unquote-regex
   "\\(\\s(\\|\\s-\\|-\\|[,%=<>\\+*/?&|!\\^~\\\\;:]\\|^\\)\\($[a-zA-Z0-9_]+\\)")
 
-(defconst julia-forloop-in-regex
-  "for +.*[^
+(defconst julia-in-regex
+  " +.*[^
 ].* \\(in\\)\\(\\s-\\|$\\)+")
 
 (defconst julia--forloop-=-regex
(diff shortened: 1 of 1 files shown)


non-Jedi:
`in` is functioning as just a normal infix function when not used in a for-loop and so shouldn't be font-locked as a keyword.

mcol:
@non-Jedi I didn't see your comment before pushing my commit. As a newbie user, I think it would help to font-lock `in` also outside a loop, but I can revert that.

non-Jedi:
My most convincing objection is that `in(x, y)` has the exact same syntactical meaning as `x in y` (it even parses the same; try it yourself with `Meta.parse`). If we font-lock `in` for `x in y`, it doesn't make sense not to font-lock it for `in(x, y)`, and if we're font-locking `in(x, y)`, it's only going to look confusing next to the non-font-locked `foo(x, y)`.

I kinda get it. Infix function symbols are parsed in a way that's different from normal symbols. `isa` is the same way. For consistency's sake, if we were going to font-lock `in` and `isa` differently we would also need to font-lock all in-fix operators (`+`, `*`, etc.).

mcol:
Sure, the more I thought of that and played with edge cases, the more I agree with you. Clearly, that makes a fix much harder, as you mentioned in the issue. I still have a couple ideas I'd like to try, but it'll be a few days before I can get back to this.

tpapp:
FWIW, I would prefer just font-locking in `for`. I realize it is not easy to do this.

non-Jedi:
@mcol, would you be interested in updating this pull request just to add the multi-loop `in` test cases and mark them as broken?

PR #173, opened 2022-04-17
Provide a safe range for julia-indent-offset

gcv:
Projects may wish to customize `julia-indent-offset` as either file-local or directory-local (`.dir-locals.el`) variables. Without the `:safe` form in `defcustom`, this causes Emacs to pop up an annoying warning. The proposed change gives a reasonable indentation range (no one should indent under 2 spaces or over 8).


tpapp (approved):
Thanks!

PR #177, opened 2022-12-21
@li1 imenu fixes now tested

non-Jedi:
Closes #124.

This takes @li1's commits from #124 and instead re-uses existing regexes currently used for font-locking which means the regexes are already tested to some extent. It also adds support for font-locking variable definition in `const` expressions.


non-Jedi:
Longer term, we should [test the indices created by imenu](https://github.com/emacs-mirror/emacs/blob/ad5a67996ddf23df904c09165475759e2e0a68b1/test/lisp/progmodes/python-tests.el#L4759), but I don't think merging this as-is should be a problem.

tpapp:
@non-Jedi, apologies for overlooking this. OK if I merge?

non-Jedi:
Hi @tpapp. Should be fine to merge. Glancing through this now, the only comment I'd make is that what's labeled as "Struct" for imenu should probably be "Type" since the regex includes abstract and primitive types.

tpapp (approved):
Thank you!

PR #181, opened 2023-01-16
Fix escaping in julia-in-string docstring

### LINKED ISSUE #180: Warning in Emacs 29
Hi!

When using Emacs 29, I receive the following warning when loading `julia-mode`:

```
⛔ Warning (comp): julia-mode.el:319:2: Warning: docstring has wrong usage of unescaped single quotes (use \= or different quoting)
```

ronisbr:
Hi!

Before this commit, the docstring of `julia-in-string` was rendered as:

Captura de Tela 2023-01-16 às 14 51 52

After, it is rendered as:

Captura de Tela 2023-01-16 às 14 52 26

This commit also suppresses a warning in Emacs 29 related to unescaped `'`:

> NEWS from Emacs 29

> *** Byte compilation will now warn about some quoting mistakes in docstrings.
> When writing code snippets that contains the "'" character (APOSTROPHE),
> that quote character has to be escaped to avoid Emacs displaying it as
> "’" (LEFT SINGLE QUOTATION MARK), which would make code examples like>
> 
>     (setq foo '(1 2 3))
> 
> invalid.  Emacs will now warn during byte compilation if it seems
> something like that, and also warn about when using RIGHT/LEFT SINGLE
> QUOTATION MARK directly.  In both these cases, if these characters
> should really be present in the docstring, they should be quoted with
> "\=".

Closes #180

### DIFF
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -329,7 +329,7 @@ Handles both single-line and multi-line comments."
 (defun julia-in-string (&optional syntax-ppss)
   "Return non-nil if point is inside a string using SYNTAX-PPSS.
 Note this is Emacs' notion of what is highlighted as a string.
-As a result, it is true inside \"foo\", `foo` and 'f'."
+As a result, it is true inside \"foo\", \\=`foo\\=` and \\='f\\='."
   (nth 3 (or syntax-ppss (syntax-ppss))))
 
 (defun julia-in-brackets ()


non-Jedi (approved):

ronisbr:
Bump! :)

tpapp (approved):
Thanks!

ronisbr:
Thanks!

PR #185, opened 2023-05-04
Use completion-at-point and abbrev-mode for LaTeX substitution, take 2

tpapp:
This finishes #100.


tpapp:
@non-Jedi, thanks for the great work on this. I understand that you did not have to time to review my changes, but I have been using this version for months now and it works fine, on Emacs 28 and 29, so I am merging. A separate PR will remove support for 25.

non-Jedi:
Thank you for pushing this over the finish line @tpapp. And apologies for ghosting you a little bit with respect to https://github.com/non-Jedi/julia-emacs/pull/1. Hopefully we won't get a sudden raft of bug reports like #188, but I'll try to at least poke my head in a bit more often to triage those that I can.

dhanak:
Hello!

I have set up my Emacs to upgrade my (m)elpa packages automatically on a weekly basis. Thus, it came as a surprise to me that tab completion substitution for LaTeX symbols in julia-mode broke some time earlier this week. Then it took me a while to investigate the issue and track the root cause down to this PR.

Granted, I have ivy-mode installed, and I noticed the comments in the source code related to [a bug](https://github.com/abo-abo/swiper/issues/2345). But that seems like an old issue, yet julia-mode worked perfectly until recently. Furthermore, there is nothing wrong with the substitution of LaTeX symbols which are unambiguous, e.g., `\circeq`, only those which have multiple possible suffixes, e.g., `\circ` and `\in`.

Moreover, when I type `\circ<TAB>` 99.99% of the time I want it to be substituted with the circle symbol, and not to be completed to some hardly ever used LaTeX symbol name that also happens to begin with `\circ`. The same goes for `\in`.

I spent the better part of an hour trying to figure out how to patch the change to bring the old behavior back, to no avail. I finally gave up and went back to the January 19 version of julia-mode which I restored from a snapshot.

Note that I don't really want to start using abbrev. As I switch back-and-forth between Emacs and REPL, I definitely want to use the same key to expand symbols, otherwise I'm sure to mess it up regularly.

I think this will be an issue for others as well, so I'd like to offer my help to come up with a solution that suits (almost) everybody.

tpapp:
@dhanak, the issue is discussed in #193. We are planning to enable something like the old behavior, and would appreciate your input there.

PR #161, opened 2021-08-22
Declare inferior-julia obsolete.

tpapp:
Cf #109, #144.

Issue #144, opened 2020-09-13
What is the purpose of inferior-julia?  Gives "Text is read only" error when typing into it.


bgoodr:
I'm referring to this function defined here:

https://github.com/JuliaEditorSupport/julia-emacs/blob/b5f5983d2b232c8bba4c5eff75cccdb787c19d98/julia-mode.el#L842

I've downloaded and installed Julia 1.5.1 into my home directory at ~/julia/julia-1.5.1/bin (Ubuntu 20.04), and have installed this package from MELPA using the construct in my Emacs setup of:

    (use-package julia-mode
      :ensure t
      :config
      ;; Set julia-program to the manually downloaded julia binary:
      (unless (boundp 'julia-program)
        (error "ASSERTION FAILED: Expected use-program to have bound julia-program by now by downloading it and installing it."))
      (let ((julia-install-dir "~/julia"))
        (when (file-directory-p julia-install-dir)
          (let ((dirs (directory-files
                       julia-install-dir
                       'full
                       (rx "julia-"
                           (+ (any "0-9" "."))
                           eol))))
            ;; Assume that only one version is installed, for now:
            (when (eq 1 (length dirs))
              (let ((julia-dir (car dirs)))
                (let ((potential-julia-program (expand-file-name "bin/julia" julia-dir)))
                  (when (file-executable-p potential-julia-program)
                    (setq julia-program potential-julia-program)))))))))

But when I execute M-x inferior-julia, I see this:

                   _
       _       _ _(_)_     |  Documentation: https://docs.julialang.org
      (_)     | (_) (_)    |
       _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
      | | | | | | |/ _` |  |
      | | |_| | | | (_| |  |  Version 1.5.1 (2020-08-25)
     _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
    |__/                   |

And when I type any character in that Emacs buffer, it gives the error:

    Text is read-only

Either this is user-error, or something is broken somehow.

I noticed that I do not see a prompt at all if I use M-x shell and then invoke the julia executable:

    $ ~/julia/julia-1.5.1/bin/julia
                   _
       _       _ _(_)_     |  Documentation: https://docs.julialang.org
      (_)     | (_) (_)    |
       _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
      | | | | | | |/ _` |  |
[...]

Issue #109, opened 2020-03-23
remove inferior-julia


tpapp:
It has been suggested in #106 to remove `inferior-julia` from the package. For those who are not familiar with it, it is a simple comint-based approach.

At the moment, we have the following approaches to interacting with the REPL (all of which are limited or WIP to a certain extent, but I will not do a comparison here):

1. [ESS-julia](https://github.com/emacs-ess/ESS/wiki/Julia)

2. [julia-repl](https://github.com/tpapp/julia-repl/)

3. [julia-snail](https://github.com/gcv/julia-snail)

4. [lsp-julia](https://github.com/gdkrmr/lsp-julia)

5. [emacs-jupyter](https://github.com/dzop/emacs-jupyter)

6. [eglot](https://discourse.julialang.org/t/using-languageserver-jl-with-eglot-in-emacs/23125)

(Sorry if I missed something, happy to add it).

The main questions are whether

1. anyone is using `inferior-julia`,
2. there is any harm in keeping it in the package, even if unused,
3. there is any benefit from removing it.

non-Jedi:
The language server protocol does not provide a REPL, so both lsp-julia and eglot can be removed. Swank.jl via SLIME does however (if  I could ever get it to work).

4aHxKzD:
1. Never used it (always used julia-repl)
2.
3. as a user, I guess the only benefit of removing it is `inferior-julia-mode` not appearing in the minibuffer when i invoke `M-x`. As a potential contributor to the package (after learning elisp...) I think that having legacy code is always unwanted.

smendozabarrera:
I use ESS-julia in org-mode, this way gives me the possibility of run smoothly the code blocks. Regards.

bgoodr:
I filed https://github.com/JuliaEditorSupport/julia-emacs/issues/144 because I tried to use it and it failed with the error I mentioned there. Perhaps it never worked, or never worked properly, so perhaps this is evidence that it should be removed.  Or if not removed, then the issue in 144 addressed somehow.

Warning: I'm a new user of this Emacs package, so my commentary may be totally off-base.

tpapp:
#144 is a good opportunity to revisit this. I would at least add a deprecation message, something like
```emacs-lisp
(make-obsolete 'inferior-julia "REPL modes are now provided by various third-party packages, this will be removed.")
```

non-Jedi:
[...]

PR #187, opened 2023-05-04
Remove inferior-julia mode.

tpapp:
Cf discussion in #109, #144, #161.

I wait for reviews or just comments until May 13, then just merge this.


tpapp:
No comments way past the date above, I am merging this.

PR #196, opened 2023-06-20
option to restore previous TAB behavior

### LINKED ISSUE #193: Consider creating an option to mimic previous tab behavior
Per @ronisbr, #185 significantly change the behavior of <kbd>TAB</kbd>.

https://github.com/JuliaEditorSupport/julia-emacs/issues/188#issuecomment-1542914710:

> If I use the latest commit with `(setq julia-force-tab-complete t)` (default), I can replace LaTeX symbols by pressing TAB. For example, `"\alpha<TAB>"` gives `α`. However, in this case, TAB always force indentation update. Hence, I am not able, for example, to increase the indentation by pressing TAB repeatedly as I could before.
> 
> If I set `(setq julia-force-tab-complete nil)`, now I can increase the indentation by pressing TAB repeatedly. However, I lost the ability to replace LaTeX symbol by hitting TAB. In this case, I must use `C-SPC`, which is my default completion keystroke.
> 
> I am not saying we must revert to the old behavior. I am just saying that the behavior has changed and I am not certain if this is a bug or how the new system must work.

To me, this brings <kbd>TAB</kbd> behavior back in line with how users of the rest of emacs would expect it to behave, and I see it as a benefit. If there's sufficient interest, I could create a configuration option that would cause <kbd>TAB</kbd> to simulate the old behavior.

tpapp:
Fixes #193.

This is preliminary, **comments are welcome**, even along the lines of "Tamás your elisp code is so horrible that I had to wash my eyes with soap after reading it, here is how you do it nicer".

Try it with
```emacs-lisp
(define-key julia-mode-map (kbd "TAB") 'julia-latexsub-or-indent)
```
Maybe we can add an option for it.

Possible improvements:

- [ ] use something else than `completing-read`? is there a nicer popup version?
- [ ] annotate with the actual replacement in `completing-read`
- [x] option for greedy replacement for a full match (eg `\bar` will replace, no popup)
- [ ] option to pick the first replacement, whatever it is

Have to do before merging this:

- [x] some unit tests
- [ ] code review from someone
- [x] some testing by interested users

### DIFF
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -938,12 +938,10 @@ buffer where the LaTeX symbol starts."
 When multiple options match, ask the user to clarify via `completing-read', unless there is a complete match and `julia-latexsub-greedy' is `t'."
   (when-let (beg (julia--latexsub-start-symbol))
     (let ((partial (buffer-substring-no-properties beg (point))))
-      (message "partial %s" partial)
       (when-let (replacements (gethash partial julia-mode--latexsubs-partials))
-        (message "replacements %s" (prin1 replacements))
         (let* ((complete-match (member partial replacements))
                (replacement (cond ((and complete-match julia-latexsub-greedy) partial)
-                                  ((cdr replacements) (gethash (completing-read "LaTeX completions: " replacements) julia-mode-latexsubs))
+                                  ((cdr replacements) (completing-read "LaTeX completions: " replacements))
                                   (t (car replacements)))))
           (cons beg replacement))))))
 
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -67,29 +67,32 @@ User can still use `abbrev-mode' or `expand-abbrev' to substitute
 unicode for LaTeX even if disabled."
   :type 'boolean)
 
+(defcustom julia-latexsub-greedy t
+  "When `t', `julia-latexsub-or-indent' does not offer options when a complete match is found. Eg for \"\\bar\", \"\\barcap\" etc will not be offered in a prompt."
+  :type 'boolean)
+
 (defconst julia-mode--latexsubs-partials
   (let ((table-unordered (make-hash-table :test 'equal))
         (table-ordered (make-hash-table :test 'equal)))
     (cl-flet ((_append (key replacement)
                        (puthash key (cons replacement (gethash key table-unordered nil)) table-unordered)))
       ;; accumulate partials
-      (maphash (lambda (latex replacement)
+      (maphash (lambda (latex _unicode)
                  (cl-assert (string= (substring latex 0 1) "\\") nil
                             "LaTeX substitution does not start with \\.")
                  (let ((len (length latex)))
                    (cl-assert (< 1 len) nil "Trivially short LaTeX subtitution")
-                   ;; for \foo, put f, fo, foo into the table
+                   ;; for \foo, put \f, \fo, \foo into the table
                    (cl-loop for i from 2 to len
-                            do (_append (substring latex 1 i) (cons latex replacement)))))
+                            do (_append (substring latex 0 i) latex))))
                julia-mode-latexsubs)
       ;; order by LaTeX part
       (maphash (lambda (partial replacements)
-                 (puthash partial (sort replacements (lambda (a b) (string< (car a) (car b)))) table-ordered))
+                 (puthash partial (sort replacements #'string<) table-ordered))
                table-unordered))
     table-ordered)
   "A hash table containing all partial strings from the LaTeX abbreviations in
-`julia-mode-latexsubs' as keys. Values are lists of the `(cons latex replacement)`,
-ordered by the `latex` part.")
+`julia-mode-latexsubs' as keys. Values are sorted lists of complete \"\\some_string\".")
 
 (defun julia-mode--latexsubs-longest-partial-end (beg)
   "Starting at `beg' (should be the  \"\\\"), return the end of the longest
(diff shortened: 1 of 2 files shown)


torfjelde:
I can be one of those "interested users":) Currently trying it out, and it seems to be working nicely :) Thanks! @tpapp !

julia-mode.el
```diff
@@ -908,6 +919,40 @@ buffer where the LaTeX symbol starts."
...
+  "When there is a LaTeX substitution that can be made before the point, return (CONS BEG LATEX).
+
+`beg' is the position of the `\`, `latex' is the string to replace, including the `\`.
+
+When multiple options match, ask the user to clarify via `completing-read', unless there is a complete match and `julia-latexsub-greedy' is `t'."
+  (when-let (beg (julia--latexsub-start-symbol))
+    (let ((partial (buffer-substring-no-properties beg (point))))
+      (message "partial %s" partial)
+      (when-let (replacements (gethash partial julia-mode--latexsubs-partials))
+        (message "replacements %s" (prin1 replacements))
```
dhanak, line 932: This is cool for debugging purposes, but it bloats the minibuffer unnecessarily, even if temporarily.

This and the previous message line should probably be deleted before merging.

julia-mode.el
```diff
@@ -908,6 +919,40 @@ buffer where the LaTeX symbol starts."
...
+        (let* ((complete-match (member partial replacements))
+               (replacement (cond ((and complete-match julia-latexsub-greedy) partial)
+                                  ((cdr replacements) (gethash (completing-read "LaTeX completions: " replacements) julia-mode-latexsubs))
```
dhanak, line 935: I think this is an error.

When the completion is not full, e.g., \cir<TAB>, I am offered a list of valid completions. When I press enter on one of them, I get a `progn: Wrong type argument: char-or-string-p, nil` error.

After some inspection, it seems that this is because this line here should return the raw latex control sequence, not the already substituted value. The substitution is performed in the  `julia-latexsub-or-indent` function.

If I change this line to

```elisp
((cdr replacements) (completing-read "LaTeX completions: " replacements))
```

then it works as expected.

dhanak (changes requested):
I tried this briefly and apart from the two comments below, it seems to work as expected. Nice work, and thanks for the effort!

dhanak:
Any chance of getting this merged soon?

tpapp:
It still needs tests. I will try to do them soon and merge your suggestions too. I appreciate the ping.

tpapp:
I incorporated the suggestions by @dhanak and added some basic tests. While one can always prettify this further, it is functional so I would prefer to merge and then leave bells & whistles for future work.

Further testing is welcome, but a code review would be best.

tpapp:
asking the usual suspects for review before merging, @giordano, @non-Jedi, @FelipeLema, @ajhalme

FelipeLema (approved):

julia-mode.el
```diff
@@ -919,6 +941,42 @@ buffer where the LaTeX symbol starts."
...
+
+When multiple options match, ask the user to clarify via `julia-latexsub-selector', unless there is a complete match and `julia-latexsub-greedy' is `t'.
+
+Presently, this is not the default. Enable with eg
+
+(define-key julia-mode-map (kbd \"TAB\") 'julia-latexsub-or-indent)
+
+eg in your `julia-mode-hook'."
+  (interactive "*i")
+  (if-let (replacement (julia-mode--latexsub-before-point))
```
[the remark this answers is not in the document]
non-Jedi: I guess this option is missing this from the OP

> option for greedy replacement for a full match (eg \bar will replace, no popup)

but I think you could handle just that case with significantly less complexity than this PR.

julia-mode.el
```diff
@@ -919,6 +941,42 @@ buffer where the LaTeX symbol starts."
...
+      (progn
+        (delete-backward-char (- (point) (car replacement)))
+        (insert (gethash (cdr replacement) julia-mode-latexsubs)))
+    (julia-indent-line)))
```
non-Jedi, line 978: Wait, if this ultimately calls `julia-indent-line`, I don't think it solves the problem from #193 at all. The issue in #193 was being able to have `TAB` work for both latex substitution as well as inserting `TAB`s (see example in this comment). I think this might do what ronisbr was looking for if it called `indent-for-tab-command` instead of `julia-indent-line`.

tpapp: Sorry, I missed this comment, just found it among messages. This PR solves the problem for me, if anyone still has problems with it please open another issue.

Specifically, it only calls `julia-indent-line` when it could not make a substitution. So, TAB tries to subtitute, and indents if it can't.  I find that behavior useful.

non-Jedi: So basically it gives you the reverse of the behavior of `indent-for-tab-command` when `tab-always-indent` is set to `'complete`:

- With `indent-for-tab-command` with `(setq tab-always-indent 'complete)`: indent then complete
- With this function: complete (but only latexsubs) then indent

If we change `julia-indent-line` in this function to `indent-for-tab-command`, the user can change between TAB inserting tabs and TAB indenting the line by customizing the value of `tab-always-indent`. I'll create a PR for your review when I have a few minutes.

If you haven't seen it yet, could you please comment on [my other review comment when you get a chance](https://github.com/JuliaEditorSupport/julia-emacs/pull/196#discussion_r1592636480)?

julia-mode.el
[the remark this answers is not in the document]
tpapp: Sorry, I do not understand. `julia-latexsub-greedy`, which is set to `t` by default, does exactly this (eg `\bar` just completes).

Regarding complexity: I agree that in an ideal world, this PR would not have been necessary. But practically, the generic completion mechanisms of Emacs have proven to be baroque and full of corner cases and/or implementation bugs like #204 (thanks for chasing many of those down!). I think that for now there is value in maintaining our own version which Just Works :tm:.

PR #197, opened 2023-07-12
Fix const definition regex to match symbols with underscores etc.

non-Jedi:
Regex introduced in #177 doesn't match e.g. `const my_symbol_with_underscores = 0` or `x = const y = 1`.

### DIFF
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -290,10 +290,10 @@ partial match for LaTeX completion, or `nil' when not applicable."
 
 (defconst julia-const-def-regex
   (rx
-   bol (zero-or-more space)
-   "const" space
-   (group (one-or-more alnum)) (zero-or-more space)
-   "=" (not (any "="))))
+   symbol-start "const" (1+ space)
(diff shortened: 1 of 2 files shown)


julia-mode.el
```diff
@@ -290,10 +290,10 @@ partial match for LaTeX completion, or `nil' when not applicable."
 
 (defconst julia-const-def-regex
   (rx
-   bol (zero-or-more space)
-   "const" space
-   (group (one-or-more alnum)) (zero-or-more space)
-   "=" (not (any "="))))
+   symbol-start "const" (1+ space)
+   (group (minimal-match (seq symbol-start (one-or-more anychar) symbol-end)))
```
tpapp, line 294:
```suggestion
   (group (minimal-match (seq symbol-start (one-or-more anything) symbol-end)))
```
because `anychar` does not work on Emacs 26.x.

tpapp (approved):
LGTM, thanks!

Issue #199, opened 2023-10-03
Test failure for julia--test-end-of-defun-nested-2


simonlegrand:
I just tried to run the tests for [7a8c868](https://github.com/JuliaEditorSupport/julia-emacs/commit/7a8c868e0d3e51ba4a2c621ee22ca9599e0e4bbb), with Emacs-29.1, and I got the following error:

```
...
Test julia--test-end-of-defun-nested-2 backtrace:
  ert-fail(((should (eq (point) (if (stringp "return fact(x)[ \n]+end"
  (if (unwind-protect (setq value-657 (apply fn-655 args-656)) (setq f
  (let (form-description-659) (if (unwind-protect (setq value-657 (app
  (let ((value-657 'ert-form-evaluation-aborted-658)) (let (form-descr
  (let* ((fn-655 #'eq) (args-656 (condition-case err (let ((signal-hoo
  (progn (julia-mode) (insert "function f(x)\nfunction fact(n)\nif n =
  (unwind-protect (progn (julia-mode) (insert "function f(x)\nfunction
  (save-current-buffer (set-buffer temp-buffer) (unwind-protect (progn
  (let ((temp-buffer (generate-new-buffer " *temp*" t))) (save-current
  (lambda nil (let ((temp-buffer (generate-new-buffer " *temp*" t))) (
  ert--run-test-internal(#s(ert--test-execution-info :test #s(ert-test
  ert-run-test(#s(ert-test :name julia--test-end-of-defun-nested-2 :do
  ert-run-or-rerun-test(#s(ert--stats :selector t :tests ... :test-map
  ert-run-tests(t #f(compiled-function (event-type &rest event-args) #
  ert-run-tests-batch(nil)
  ert-run-tests-batch-and-exit()
  command-line-1(("-L" "." "-l" "ert" "-l" "julia-mode-tests.el" "-f" 
  command-line()
  normal-top-level()
Test julia--test-end-of-defun-nested-2 condition:
    (ert-test-failed
     ((should
       (eq
	(point)
	(if ... ... "return fact(x)[ \n]+end")))
      :form
      (eq 141 163)
      :value nil))
   FAILED  19/65  julia--test-end-of-defun-nested-2 (0.001256 sec) at julia-mode-tests.el:887
...
```
I tried with Emacs-28.2 and it works fine.

Cheers,

Simon

rdrg109:
This problem seems to be the cause of the problem  for this behavior in Guix

```
[...]

PR #202, opened 2024-01-19
Fix highlighting for macros prefixed by "!"

### LINKED ISSUE #201: Macro invocations not highlighted if preceded by an exclamation mark
`julia-model` doesn't recognise macro invocations when they are preceded by an exclamation mark, such as when negating a boolean expression.

```julia
if @foo()  # <- macro invocation is highlighted
    return 1
end
if !@foo()  # <- macro invocation is not highlighted
    return 0
end
```

non-Jedi:
Closes #201.

Checked other regexes for similar issues involving "!" and did not discover any of significance. The Julia manual specifically calls out "!" as being special in this way and no other symbols, so just including "!" as a special-case in regexes is sufficient.

Test failure on emacs 29 is unrelated and due to #199.

### DIFF
--- a/julia-mode-tests.el
+++ b/julia-mode-tests.el
@@ -768,6 +768,13 @@ var = func(begin
     (julia--should-font-lock string 22 nil) ; =
     ))
 
+(ert-deftest julia--test-!-font-lock ()
+  (let ((string "!@macro foo()"))
+    (julia--should-font-lock string 1 nil)
+    (julia--should-font-lock string 2 'julia-macro-face)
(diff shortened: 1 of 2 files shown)


ViralBShah:
@non-Jedi I've invited you to be a member of this organization. Please merge the PR if you think it is good to go.

non-Jedi:
Thanks. Would like to have at least one other person review before merging. Pinging @tpapp, @giordano, @FelipeLema: could one of you take a quick look?

FelipeLema (approved):

PR #206, opened 2024-03-15
Strip whitespace in latexsub :exit-function

### LINKED ISSUE #204: Unicode substitution broken with helm
@giordano [reports](https://github.com/JuliaEditorSupport/julia-emacs/issues/193#issuecomment-1909226438):

>Yes, "\alph" for me is completed to "\alpha", not "α", further pressing `M-C-i` does nothing. Not sure it's relevant, but I use Helm. I also have auto-complete installed, but doesn't seem the mode is active in my julia buffer.

Locally, I've confirmed that with only `julia-mode` installed, `C-M-i` correctly completes "\alph" to "α". If helm is installed with the following minimal config, `C-M-i` completes "\alph" to "\alpha":

```elisp
(require 'package)
(add-to-list 'package-archives '("melpa" . "https://melpa.org/packages/") t)
(package-initialize)

(package-install 'julia-mode)

(package-install 'helm)
(helm-mode 1)
```

This seems like a re-occurrence of https://github.com/emacs-helm/helm/issues/2356, but more investigation is needed.

non-Jedi:
According to the docstring for `completion-extra-properties`, the `string` passed to the `:exit-function` should be the bare text to which the field was completed, but helm-mode adds an extra space at the end of `string`. Since none of our latexsubs include whitespace, we can safely strip whitespace in the `:exit-function` and work around this helm bug.

I will file an upstring bug report against helm, but I think it's worth fixing quickly here since it's only a simple 3-line change.

Closes #204.

@tpapp, @giordano, or @FelipeLema could one of you review please? This fixes an extremely annoying bug for anyone using helm. @giordano, could you please double-check that this fixes the issue for you if you get a chance?

### DIFF
--- a/julia-mode-tests.el
+++ b/julia-mode-tests.el
@@ -963,6 +963,8 @@ end" 'end-of-defun "n == 0" "return fact(x)[ \n]+end" 'end 2))
   (should (equal (julia--call-latexsub-exit-function
                   "\\kappa\\alpha(" 7 13 "\\alpha" t)
                  "\\kappaα("))
+  ;; Test that whitespace is stripped from `:exit-function' NAME for compatibility with helm
+  (should (equal (julia--call-latexsub-exit-function "x\\alpha " 2 8 "\\alpha " t)  "xα "))
   ;; test that LaTeX not expanded when `julia-automatic-latexsub' is nil
   (should (equal (julia--call-latexsub-exit-function "\\alpha" 1 7 "\\alpha" nil) "\\alpha"))
   (should (equal (julia--call-latexsub-exit-function "x\\alpha " 2 8 "\\alpha" nil)  "x\\alpha "))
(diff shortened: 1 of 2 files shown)


non-Jedi:
For anyone curious on history, this is the same bug as https://github.com/emacs-helm/helm/issues/2360. I'm not at all sure how I had helm working before given that Thierry only fixed that issue for insertion but not for what `:exit-function` sees. Based on what I see there, an upstream solution to this bug seems unlikely, and this workaround is relatively painless.

tpapp (approved):
LGTM, thanks!

FelipeLema (approved):

non-Jedi:
Thanks @tpapp and @FelipeLema! Merged a 2dfc869ff6b3878407fe7226669dfaae8d38d541.

giordano:
Sorry, I had a chance to test this only now!  It mostly works in the sense that `\alph<TAB>` is replaced with `α` straight away (yay!), but `\alpha<TAB>` does nothing, `\alpha` remains there.  Thanks for looking into this!

non-Jedi:
Looks like helm-mode isn't calling the `:exit-function` when the symbol is already complete. I don't immediately see a way of working around this without an upstream fix. Will open an issue against helm once I've created a minimal reproducer.

PR #207, opened 2024-04-09
Remove unused cl lib

rtpg:
By my reading of this source file there's no actual usage of the stuff in cl anymore so we don't even need to use cl-lib (I might be wrong here though)


tpapp (approved):
Thanks!

PR #208, opened 2024-04-09
add 29.x Emacs versions to CI matrix

tpapp:


tpapp:
Currently the nested defun is broken (finds the inner one). I would wait to fix that in an issue and add CI regardless.

non-Jedi (approved):
Looks good to me. Definitely a positive change to factor that hairy macro interpolation into a function. Note that the test failure on 29 already has an issue: #199

tpapp:
Thanks. I will look into #199 first before making an exception CI.

PR #209, opened 2024-04-09
Change tests so that we test our own implementations only.

tpapp:
Bugs in wrappers do not break our code tests then. Cf #199.

### DIFF
--- a/julia-mode-tests.el
+++ b/julia-mode-tests.el
@@ -80,7 +80,7 @@
            (if (stringp to)
                (progn (goto-char (point-min))
                       (re-search-forward to)
-                      (if end
+                      (when end
                           (goto-char (match-end 0))
                         (goto-char (match-beginning 0))
                         (point-at-bol)))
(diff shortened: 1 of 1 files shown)


tpapp:
asking the usual suspects for review before merging, @giordano, @non-Jedi, @FelipeLema, @ajhalme

julia-mode-tests.el
```diff
@@ -74,25 +74,26 @@
...
       (goto-char from))
-    (funcall fun arg)
+    (apply fun args)
     (let ((actual-to (point))
           (expected-to
            (if (stringp to)
                (progn (goto-char (point-min))
                       (re-search-forward to)
-                      (if end (goto-char (match-end 0))
+                      (if end
```
FelipeLema, line 83:
```suggestion
                      (when end
```

FelipeLema (approved):
minor comment, still 👍🏾

tpapp:
I was too hasty to accept that suggestion, it is a multi-branch `if`, so `when` is not a substitute. Thanks for the review!

PR #210, opened 2024-04-09
run the latex substitutions update script.

tpapp:


tpapp:
CI failures are #199.

PR #213, opened 2024-09-25
Release 1.0, clarify SemVer, add use-package example

### LINKED ISSUE #212: `julia-last-open-block-pos` broken on latest version on NonGNU ELPA (0.4)
Both `melpa-stable` and `NonGNU` ELPA are at version `0.4`.  However, the two
versions, while nominally being the same, differ in functionality.  The version
on `melpa-stable` corresponds to the git tag which references commit
8bfc709716a257521cb386f20b8932e83db930a9 dated `2020-03-09`.  Whereas the
version on `nongnu` ELPA correspond to
6408b96c1c97e41bc2af060d661afee4f7b22e89 dated `2019-11-13` which is when the
`Version` in the package header first was bumped to `0.4`.  Interestingly, this
commit also introduced a bug in `julia-last-open-block-pos` as a result of an
"over-enthusiastic" renaming.  The result of this bug is that pressing `TAB` to
indent in blocks is broken.

Now, the very next commit (9df3a7ec32f3587dd3a590ae37535fdea6e88a0d) on
`2019-11-13` fixed the bug.  On `2020-03-09` this commit was
released via the `0.4` tag and made its way into `melpa-stable`.  Unfortunately,
since the version in package header wasn't updated, this commit (and subsequent
ones) never made their way into `nongnu` ELPA.

## Proposal

Given the peculiarities of `nongnu` (and `gnu`) ELPA releases, I propose that a
version `0.5` be created, with the earliest commit where the version in the
package header is updated to `0.5`, being the one that is tagged with `0.5`.
Doing so will bring `melpa-stable` and `nongnu` ELPA releases in sync.

Thereafter, (in the very next commit) I propose that the version in package
header be updated to `0.5.50-git` (assuming that you expect the next release to
be either `0.6` or `1.0`) or `0.5.0.50-git` (if you'd like to keep the option
open for the next release to be `0.5.1`).  The `-git` suffix will be ignored by
`nongnu` ELPA.  Once you're ready to release the subsequent version, you can
once again make a commit to bump the version in the package header, and add a
git tag (ensuring that `melpa-stable` and `nongnu` version continue to be
synchronized).

## References

See [this comment](https://github.com/melpa/melpa/issues/2955#issuecomment-1519154736)
and associated discussion in case you'd like to know more about ELPA versioning.

tpapp:
Fixes #212. From now on we take semantic versioning seriously.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,17 @@ Unless you want to develop this package, it is recommended that you use it from 
 (require 'julia-mode)
 ```
 
+Alternatively, if you are using [`use-package`](https://github.com/jwiegley/use-package), which has been [part of Emacs](https://www.gnu.org/software/emacs/manual/html_mono/use-package.html) since version 29.1, you can use
+
+```elisp
+(use-package julia-mode
+  :ensure t)
+```
+
+### A note on versioning
+
+The code has been on a “rolling release” model before version 1.0: not all code changes were accompanied by a version increment. From version 1.0, we follow [semantic versioning](https://semver.org/).
+
 ### Installing from Source
 
 To get the latest version of `julia-mode`, clone this repository and then use:
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -1,8 +1,8 @@
 ;;; julia-mode.el --- Major mode for editing Julia source code -*- lexical-binding: t -*-
 
-;; Copyright (C) 2009-2014 Julia contributors, 2015-2020 julia-mode contributors
+;; Copyright (C) 2009-2014 Julia contributors, 2015-2024 julia-mode.el contributors
 ;; URL: https://github.com/JuliaEditorSupport/julia-emacs
-;; Version: 0.4
+;; Version: 1.0.0
 ;; Keywords: languages
 ;; Package-Requires: ((emacs "26.1"))
 


FelipeLema (approved):
🚀

giordano (approved):

tpapp:
@suhail-singh: sorry I don't understand. Should I merge master into this branch? Is this process written down somewhere?

What difference does it make? I was under the impression that MELPA scans daily and picks up what we have.

tpapp:
@suhail-singh: thanks for the information. I understand that *technically*, any given version would point to different commits, but it would still be the same code, right? So I don't see that as a problem.

Can't speak for the others, but there is a limited amount of crazy I am willing to tolerate to support every repo scheme. We can just **have this package removed from MELPA stable if that becomes problematic**. 

Cf https://github.com/melpa/melpa/issues/6656. @alphapapa, since you kindly commented on #212, I would appreciate your advice here.

alphapapa:
> Cf [melpa/melpa#6656](https://github.com/melpa/melpa/issues/6656). @alphapapa, since you kindly commented on #212, I would appreciate your advice here.

Hi Tamas,

I think Suhail is correct: it would seem a bit unusual for the commit with a version tag to not be the same commit in which the version header was set to the same value, but as you said, in this case, and especially for this first 1.0.0 release, it shouldn't make any difference, since the code is the same.

I use a similar scheme for my packages: After releasing a new version, I increment the Y number (as in X.Y.Z), adding a `-pre` suffix (which is accepted by Emacs's version-number functions) to the header and changelog.  Then when I release that version, I remove the suffix and tag that commit with the matching `vX.Y.Z` tag.  Much of this comes down to personal preference.

IME the most important thing to keep in mind is that, while MELPA just builds whatever you push, effectively cloning the repo anew each time it builds a package, GNU ELPA *pulls* changes from the repo, and it will refuse to build if it pulls non-fast-forward changes (in which case it will send the maintainer an email, and you'll have to do a slightly awkward merge to resolve it).  

So if your package is built on GNU ELPA, it's important to never rewrite history on the branch that ELPA builds from.  If you make a mistake and realize it after pushing, you should just push a new commit on top that fixes the problem; and if that means incrementing the version number again (e.g. adding a `.Y` bugfix number), so be it.  (Technically you can sometimes get away with it, but it's a gamble that ELPA hasn't pulled the mistaken commit yet.)

Another thing I can recommend is to have ELPA pull from a dedicated branch rather than `master`, which will allow you to maintain a stable branch with bugfixes, apart from `master`'s being where the next `X.Y` version is developed (e.g. after releasing a bugfix to the stable branch, that branch can be merged into `master`).  But this is a matter of preference.

I hope this helps.  It can be a bit confusing at first, but once you grok the different systems, it's not too bad.  :)

tpapp:
Thanks everyone for the detailed information. I recognize that there is a "right" way of doing things, but I have to explore the tooling about that, so for now I will just merge this commit as is, and we will stick to SemVer. I think it is already an improvement.

cstjean:
Out of curiosity, what is SemVer for such a package? That the key bindings won't change in a minor version?

tpapp:
Good question, my intention is mostly to play it by ear and open a discussion if there is any doubt. But generally, bug fixes increment the patch version, features increment the minor version, and if the functions we expose change or are removed, we bump the major version. I would not include keybindings and defaults, maybe we should clarify that.

meedstrom:
Perhaps [EffVer](https://jacobtomlinson.dev/effver/) is more intuitive?

@suhail-singh out of curiosity, do you routinely look at Emacs package version strings and change upgrade decisions accordingly, or do you know anyone who does? I'd love to see a YouTube video demonstrating how to do that. 

I currently have the impression that, for a packager, tending to the exact semantics is a bit like alphabetizing bookshelves.  Basically, just do it if it's fun. Once we have a featureful package manager, with lockfiles, dependency resolution, and ability to downgrade, that situation might change.

meedstrom:
Thank you for describing your process. It's nice that the version info benefits somebody. Sorry for hijacking the thread.

Issue #136, opened 2020-07-17
Give internal functions julia-- prefix


non-Jedi:
At the next major version bump, we should rename all internal functions with `julia-` prefix to have a `julia--` prefix. Here's the list I identified after a quick survey. Please reply if you believe any of these should be part of the packages public interface. Many of these should also be given a `-p` predicate suffix.

- `julia-syntax-comment-or-string-p`
- `julia-in-comment`
- `julia-in-string`
- `julia-in-brackets`
- `julia-at-keyword`
- `julia-safe-backward-sexp`
- `julia-following-import-export-using`
- `julia-last-open-block-pos`
- `julia-last-open-block`
- `julia-paren-indent`
- `julia-prev-line-skip-blank-or-comment`
- `julia-indent-hanging`
- `julia-indent-in-string`
- `julia-indent-import-export-using`
- `julia-syntax-context-type`
- `julia-looking-at-beginning-of-defun`
- `julia-beginning-of-defun`
- `julia-end-of-defun`
- `julia-looking-at-beginning-of-defun`

Cross-reference https://github.com/JuliaEditorSupport/julia-emacs/pull/115#discussion_r454986503

tpapp:
Thanks for making this list, I agree with both the contents and adding `-p` where applicable.

PR #215, opened 2024-11-28
Update julia-mode.el

kevinli1993:
Make `julia-beginning-of-defun` a command

### DIFF
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -782,6 +782,7 @@ With positive ARG search backwards else search forward.
 ARG nil or 0 defaults to 1.  When searching backwards,
 nested defuns are handled depending on current point position.
 Return non-nil (point) if point moved to `beginning-of-defun'."
+  (interactive)	
   (when (or (null arg) (= arg 0)) (setq arg 1))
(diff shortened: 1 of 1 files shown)


tpapp:
Can you please explain why you need this? This is a function meant to be called by Emacs Lisp code under the usual circumstances.

non-Jedi:
I think the proper fix is to make `julia-end-of-defun` non-interactive. I've added a note for this to #136.

Could you do a little digging to see if there's any rationale for why you might want `end-of-defun-function` to be interactive but not `beginning-of-defun-function`? I noticed when I was poking around that `python-nav-end-of-defun` is interactive while `python-nav-beginning-of-defun` is not; didn't check any  other language major modes.

tpapp:
My hunch is that @nverno introduced this in #63 basing the code on the Python mode, and there is no particular reason for it in our codebase. @nverno, any comments?

I am inclined to just remove this. And then address #136, now that we tag versions again :wink:

tpapp:
Closing this in favor making `...end...` version internal, as listed in #136.

PR #217, opened 2024-12-13
Add `public` keyword

emmt:
This PR adds the `public` keyword that appears in Julia ≥ 1.11 with the same semantics as `export`.


tpapp (approved):
LGTM, thanks! I don't think we need separate tests for this as it uses the same logic as `export`, which we test for.

tpapp:
@non-Jedi @giordano @FelipeLema @ajhalme 

if one of you could please provide an additional review, I would merge

giordano (approved):

FelipeLema (approved):
🚀

PR #218, opened 2024-12-16
Fix error when trying to expand latex symbol at beginning of buffer

dhanak:
With company-mode, the following error is thrown when typing (anything but a latex sequence) at the start of the buffer:

```
Debugger entered--Lisp error: (wrong-type-argument number-or-marker-p nil)
  =(92 nil)
  (if (= 92 (char-before)) (progn (- (point) 1)))
  (save-excursion (while (not (or (bobp) (= 92 (char-before)) (member (char-syntax (char-before)) '(32 60 62 92)))) (backward-char)) (if (= 92 (char-before)) (progn (- (point) 1))))
  julia--latexsub-start-symbol()
  julia-mode-latexsub-completion-at-point-around()
  completion--capf-wrapper(julia-mode-latexsub-completion-at-point-around optimist)
  company--capf-wrapper(julia-mode-latexsub-completion-at-point-around optimist)
```

This change fixes that.

### DIFF
--- a/julia-mode.el
+++ b/julia-mode.el
@@ -187,7 +187,7 @@ partial match for LaTeX completion, or `nil' when not applicable."
         (group "'")        ; start single quote of character constant
         (or                ; two alternatives
          (not (any "\\"))  ; one character, not backslash
-	     (seq "\\"         ; sequence of a backslash followed by ...
+         (seq "\\"         ; sequence of a backslash followed by ...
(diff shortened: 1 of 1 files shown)


tpapp:
I ran into this too and just noticed your PR when submitting mine. I am closing it in favor of #219, which has the same fix but adds a test.

PR #219, opened 2025-03-04
Fix julia--find-latex erroring when called at the beginning of buffer.

tpapp:
Add a test for this case (which failed previously).


tpapp:
Trivial PR fixing a reasonably clearcut case, but nevertheless I would appreciate a review from someone.

PR #221, opened 2025-04-07
Dust off the README. Fixes #216.

### LINKED ISSUE #216: document package features in the README
Our README is currently a bit sparse and technical. We should expand on it by explaining some package features.

Specifically:

### highlights from the API

Everything we define with `defcustom` should be mentioned. Various customizations, such as `(define-key julia-mode-map (kbd "TAB") 'julia-latexsub-or-indent)` to get the old tab completion, should be mentioned.

### related packages

julia-repl, eglot-jl, julia-ts-mode, other suggestions welcome

### maybe

Maybe a screenshot? Eg as embedded HTML.

tpapp:


tpapp:
Reviews and additional suggestions are welcome; unless someone objects I plan to merge this tomorrow and bump the versions. We can always add additional links later.

non-Jedi (approved):

